PR #8, opened 2016-08-29
Display a list of mobile-friendly projects

### LINKED ISSUE #5: Display list of projects on home page
List should be curated to include only mobile-friendly projects

robinschaaf:
*  Clean up android js
  *  Add constants for the Project list (uses a different list of IDs for production and development)
  *  Add new components for the ProjectList and individual Project listing

Closes #5

### DIFF
--- a/Project.js
+++ b/Project.js
@@ -1,0 +1,50 @@
+import React, { Component } from 'react';
+import {
+  Linking,
+  StyleSheet,
+  Text,
+  TouchableOpacity,
+  View
+} from 'react-native';
+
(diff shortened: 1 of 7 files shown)


wgranger:
Looks pretty straightforward at this point. I wonder if there's a way we can implement NPM to run a testing localhost for native apps using the simulator. That would make testing these PRs/branches more efficient. All of the links are taking me to the Zooniverse webpage at this point, but I assume that's because the `slug` property hasn't been utilized yet.

PR #9, opened 2016-09-01
Add avatar and URL to Project List

robinschaaf:
*  Modify Project component to display the avatar and link to the project URL
  *  Add use of Orientation detection (to lock to Portrait - may wish to revisit later, but redrawing of UI upon orientation change is buggy right now)
  *  Some styling updates resulting from adding the avatar


wgranger:
Yes, so a few instances of something we might want to ignore. Perhaps `jsbundle.meta` is something we want to place in gitignore? I'm not sure we want to commit stuff like that to the repo. I'm also curious about `project.pbxproj`.

robinschaaf:
OK - It looks like I hadn't committed the gitignore before so that's in there now :-)  Also I removed those files - let me know if it breaks anything on your end.

wgranger:
Yep. That did it. Guess you can't ignore the `pbxproj` file.

PR #16, opened 2016-09-06
Add eslint and clean up files

robinschaaf:
Add eslint and lint files.  There are still a couple of violations that will be cleaned up - The console errors will be cleaned up with a subsequent branch (add-error-messaging) and the new variables added for styling will be consolidated once react-native-extended-stylesheet is added.


wgranger:
LGTM

PR #17, opened 2016-09-06
Add network detection and error for iOS and Android

### LINKED ISSUE #11: Test for network connection
Since the project display pulls avatars, etc. from the internet we need to fail gracefully in case the user doesn't have an internet connection (I believe this is also tested during the iTunes approval process)

robinschaaf:
Add network detection and an error in case a network connection loss is detected.
Probably the easiest way to test this is in the iOS simulator is by turning off your computer's wi-fi, in android you have to go into settings and turn cellular off.
Also converts ProjectList from class to react component

Closes #11

### DIFF
--- a/ProjectList.js
+++ b/ProjectList.js
@@ -2,11 +2,11 @@ import React from 'react'
 import {
   Dimensions,
   ListView,
+  NetInfo,
   StyleSheet,
   Text,
   View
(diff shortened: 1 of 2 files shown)


robinschaaf:
Will reopen against branch on my repo

PR #20, opened 2016-09-06
Convert console errors to alerts

### LINKED ISSUE #12: Add error messaging
Test that the catches from APIs are being executed (e.g. in case a Project ID is no longer valid, it is currently throwing an error) and add error messaging.

robinschaaf:
Add popup alert messages when errors occur.  Converts the two using console statements into alerts.

Closes #12

### DIFF
--- a/ProjectList.js
+++ b/ProjectList.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import {
+  Alert,
   Dimensions,
   ListView,
   StyleSheet,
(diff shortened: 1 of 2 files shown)


ProjectList.js
```diff
@@ -43,7 +44,10 @@ var ProjectList = React.createClass({
         });
       })
       .catch((error) => {
-        console.error(error);
+        Alert.alert(
+          'Error',
+          'The following error occurred.  Please close down Zooniverse and try again.  If it persists please notify us.  \n\n' + error,
```
wgranger: `\n\n` is needed twice here?

robinschaaf: Yeah it's just twice for extra space, I think it reads a slightly better (and btw, if you have any ideas for wording changes, feel free to add them!)

image

vs. 

image

wgranger:
Makes sense. Good to see via the screenshots you were able to test out these errors.

PR #23, opened 2016-09-07
Add google-analytics-bridge

robinschaaf:
Implement google analytics on the "Zooniverse Mobile" GA property.  This PR adds tracking of "Home" views and tracking of "View Project" events (with project name as the event action).

To verify this PR works you'll need access to our google analytics - under Zooniverse account -> Zooniverse Mobile -> Real-time  (Laura can give you access if you don't' have it, I can show you how to view the reports.

### DIFF
--- a/Project.js
+++ b/Project.js
@@ -9,9 +9,12 @@ import {
   View
 } from 'react-native';
 import EStyleSheet from 'react-native-extended-stylesheet'
+import GoogleAnalytics from 'react-native-google-analytics-bridge'
 
 class Project extends Component {
   handleClick() {
+    GoogleAnalytics.trackEvent('View Project', this.props.project.display_name);
+
     const zurl=`http://zooniverse.org/projects/${this.props.project.slug}`
     Linking.canOpenURL(zurl).then(supported => {
       if (supported) {
--- a/Project.js
+++ b/Project.js
@@ -13,7 +13,7 @@ import GoogleAnalytics from 'react-native-google-analytics-bridge'
 
 class Project extends Component {
   handleClick() {
-    GoogleAnalytics.trackEvent('View Project', this.props.project.display_name);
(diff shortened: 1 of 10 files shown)


Project.js
```diff
@@ -9,9 +9,12 @@ import {
   View
 } from 'react-native';
 import EStyleSheet from 'react-native-extended-stylesheet'
+import GoogleAnalytics from 'react-native-google-analytics-bridge'
 
 class Project extends Component {
   handleClick() {
+    GoogleAnalytics.trackEvent('View Project', this.props.project.display_name);
```
wgranger: Should we make this camel case? Might want to check out how we currently track on GA to stay consistent.

index.android.js
```diff
@@ -5,8 +5,12 @@ import {
 } from 'react-native';
 import EStyleSheet from 'react-native-extended-stylesheet'
 import ProjectList from './ProjectList'
+import GoogleAnalytics from 'react-native-google-analytics-bridge'
 import theme from './theme'
 
+GoogleAnalytics.setTrackerId('UA-1224199-62')
```
wgranger: Can this tracker ID be hardcoded like this?

Project.js
robinschaaf: Are you talking about the 'View Project'?  It looks like right now we don't have it very consistent on zooniverse global, but we can definitely define how we want this to look.  I'm fine with making it camelcase (just want to verify that's what you're talking about)

image

index.android.js
robinschaaf: Normally I think this would preferably use an environment variable, but unfortunately that's not available in a mobile context.  What if I just pulled it out to define in a separate file so it's at least abstracted out?

wgranger: Sounds good.

robinschaaf: OK, just changed, let me know what you think of that solution.

PR #26, opened 2016-09-09
Google analytics - change screen tracking to Event (on Android)

robinschaaf:
Make sure the tracking for Android mimics the tracking for iOS


alexbfree:
Just wanted to note that there is an event called "view" posted to Google Analytics via the Geordi client when a new subject is displayed for classification (all Geordi events are also posted to GA). Therefore you might want to use a different event make such as "page-view" if this is something other than viewing a subject to classify.

alexbfree:
Also, we already load Google Analytics via Google Tag Manager, so be careful that due to the import in this file we don't load GA twice.

PR #28, opened 2016-09-21
Add Pusher Notification Support for iOS

robinschaaf:
*  Adds cocoapods
  *  Installs Pusher pod
  *  Add support to iOS for pusher notifications

About adding the CocoaPods files to git - Cocoapods recommends keeping them in source:
http://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control

***  Note that now to open the project in xcode you should use .xcworkspace instead!! ***


wgranger:
It looks like 99% of this PR is from the pusher library. Since their (libPusher) `push-notifications` branch is still in Beta, there's a good chance we'll have to adjust this code if they make any changes. Also, I noticed you set CocoaPods to accept ios 9.0 as the oldest acceptable version. Any reason for this? Seems like some people might still be running 8.0, for example, even considering 10.0 came out the other day.

wgranger:
Also, I'm now getting `module 'HockeySDK' not found` when I try to run ios. Perhaps I'm missing something here.

robinschaaf:
Hmm, no reason to set the ios to 9.0 (besides that was the default), I'll test setting it to something older and see if it still works.

Re: HockeySDK - This was something I'm not sure should be checked in (again it's an external library).  I spoke to Chelsea about it and she said not to check it in - but as you've noticed it makes it so you can't run it by just checking it out :-(  You may want to ping Amy, she brought this up when she merged that branch and I sent her the HockeySDK install instructions, but I'm not sure if she was able to get it to work.
If we feel like checking those in are OK, then I'm happy to add them. (after all, we're checking in all this pod code)

robinschaaf:
OK, this is working for iOS 10 now, I also removed the requirement of the version 9.0 from the Podfile and re-ran and it seems to work OK this way now.  I've removed the WIP.

wgranger:
This looks good, but we might need to rethink checking in HockeyApp, per our conversation this morning.

PR #29, opened 2016-09-23
Add simple Redux implementation, Ramda

robinschaaf:
Wire up a simple Redux implementation, with a dummy setUser action (this will be add to and/or changed later, this is just to verify it's working)
Also includes adding Ramda to help with dealing with immutable state in redux (I'm ok with changing this to a different library if desired - just thought it may be fun to work with more functional stuff)


wgranger:
I assume this is ready for a review and merge.

robinschaaf:
Sure, anytime is fine.

rogerhutchings:
Looks good - there are a couple of things that aren't used (there are no actions used in the main app component, so no need to use `mapDispatchToProps`), but as there are other open PRs, we'll get them merged in and review afterwards.

PR #30, opened 2016-10-14
Add login router styling

robinschaaf:
Note: some of these differences should resolve once PR #29 is in.
Styling updates:
-  Add fonts and iconsets - Open Sans, Zooniverse Icons, Font Awesome
-  Update Android assets/images and app label (not the full name - just the app label only for now)
-  Update the iOS launch screen
-  Add spinner overlay
-  Add high level category page (and rename to Disciplines to be consistent with PFE)
  App additions:
-  Add login screen
-  Add router
-  Add simple store for storing user info locally


robinschaaf:
I'll get these merge conflicts cleaned up and push back up

src/components/Discipline.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+});
+
+Discipline.propTypes = {
+  icon: React.PropTypes.string.isRequired,
+  title: React.PropTypes.string.isRequired,
+  tag: React.PropTypes.string.isRequired,
+  color: React.PropTypes.string.isRequired
+}
+
```
wgranger: Is there a reason these propTypes are defined here rather than above in the Discipline class?

wgranger:
Is it common practice in React Native to put styles in file with the component? Just curious, as I've noticed that's how we've been doing it.

src/components/ProjectList.js
```diff
@@ -66,7 +41,7 @@ class ProjectList extends React.Component {
   }
 
   getProjects(){
-    apiClient.type('projects').get({id: MOBILE_PROJECTS, cards: true, sort: 'display_name'})
+    apiClient.type('projects').get({id: MOBILE_PROJECTS, cards: true, tags: this.props.tag, sort: 'display_name'})
```
wgranger: What does this `tags` key request? I'm unfamiliar with this.

robinschaaf: TBH, I looked up how the Projects/Discipline filter works in PFE, and that is how they query / filter on Disciplines in the project-card-list component.  I'm not sure why it has that name, but I'm guessing that disciplines are "tagged" on the project.

robinschaaf:
Re: styles in component:  That's how all of the sample code and sample projects I've seen do it handle it.  It is possible to have them pulled out in a separate file, but I've personally found it easier to have it contained together.

PR #54, opened 2016-11-14
Add Publications screen to side nav bar

robinschaaf:
Add new publication screen with a "Category" (Discipline) filter.
We should eventually separate this so that we're not duplicating effort when adding new disciplines.  (Currently this is also hardcoded in PFE)


src/actions/index.js
```diff
@@ -118,3 +123,29 @@ export function fetchProjects(parms) {
...
+
+export function fetchPublications() {
+  return dispatch => {
+    map((key) => {
+      addIndex(forEach)(
+        (project, idx) => {
+          dispatch(setState(`publications.${key}.projects.${idx}.publications`, project.publications))
+          dispatch(setState(`publications.${key}.projects.${idx}.slug`, project.slug))
+
+          if (project.slug) {
```
wgranger: Is this supposed to fetch a link and avatar for each project under publications? Doesn't look like the projects are linked.

src/components/Publication.js
```diff
@@ -0,0 +1,73 @@
...
+  render() {
+    return (
+      <View style={styles.container}>
+        <View style={styles.titleContainer}>
+          <StyledText textStyle={'citation'} text={ this.props.publication.citation } />
+            <TouchableOpacity
+              activeOpacity={0.5}
+              onPress={this.handlePress}>
+              <Icon name="angle-right" style={styles.icon} />
+            </TouchableOpacity>
```
wgranger: Is it possible to make the entire citation clickable? Upon first view, and seeing the `angle-right` icon, I assumed the entire row/citation was clickable as a link to the article. Thoughts?

wgranger:
Made some line comments. Also, I'm unable to "view all" publications on the page. It only shows Galaxy Zoo publications when I try this.

src/components/Publication.js
robinschaaf: Yep good idea!

src/actions/index.js
robinschaaf: This uses the link (project slug) to fetch the avatar and display name of each project.  the project slug is hardcoded along with each of the publications. (I hope that answers the question?)

robinschaaf:
Hmm, what device/emulator is this on?  I have to scroll down about half way before seeing others, we do have a lot of galaxy zoo ones, but it shows them all on both my android and ios emulators.

wgranger:
Oh wait, disregard that last comment. I've been in PFE land for too long and was trying to scroll with the two-finger gesture. View all is working as expected.

src/actions/index.js
wgranger: I see. I think this should be fine for this PR.

wgranger:
Ok. Looks good.

PR #55, opened 2016-11-14
Add About screen

robinschaaf:
Add new About screen with the current text from PFE (this may change after feedback)


wgranger:
LGTM

PR #56, opened 2016-11-14
Allow users to continue as guest

robinschaaf:
Adds an option on the sign in screen to "Continue without signing in"
  *  Adds "isGuestUser" boolean to user in the redux store
  *  Update the UI in a couple of places this affects - display "Guest User" rather than the user display name, alter the side drawer to give an option to "Sign In" and remove the "Sign Out" option
  *  Move around the order of the buttons on the sign in screen a bit

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -64,6 +64,15 @@ export function setUserFromStore() {
   }
 }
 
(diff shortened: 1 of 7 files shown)


wgranger:
Looks good.

PR #57, opened 2016-11-14
Fix android permissions

### LINKED ISSUE #47: Android asking for too many permissions
Android is asking for permission to View Files on devices, but it shouldn't be.  Works fine on mine but it showed up on Grant's.  It may be fixable with this solution:
https://github.com/facebook/react-native/issues/5886

robinschaaf:
Removes unnecessary Android permissions being automatically added by RN.

Changes SYSTEM_ALERT_WINDOW to only display in debug mode
Removes BLUETOOTH, READ_EXTERNAL_STORAGE and READ_PHONE_STATE 

You can check the permission changes either through the device or emulators settings, or by following the instructions here (requires an apk and adb in your path):  http://stackoverflow.com/questions/21091022/listing-permissions-of-android-application-via-adb
Fixes #47

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -1,11 +1,15 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.zooniversemobile"
     android:versionCode="1"
-    android:versionName="1.0">
+    android:versionName="1.0"
+    xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" tools:remove="${excludeSystemAlertWindowPermission}"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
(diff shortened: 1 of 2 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -1,11 +1,15 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.zooniversemobile"
     android:versionCode="1"
-    android:versionName="1.0">
+    android:versionName="1.0"
+    xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" tools:remove="${excludeSystemAlertWindowPermission}"/>
```
wgranger: So this line, in effect, will only be executed in the release version?

robinschaaf: Yes, the "remove" part will only happen in release - so that the SYSTEM_ALERT_WINDOW is only for debug.  I can rewrite this to make more sense, since it's a double-negative now?

wgranger: Ahh. Just merged this since we discussed it. If you can find a more simple and readable solution, I think it's worth doing for another PR.

PR #63, opened 2016-11-17
Refactor project list display a bit

### LINKED ISSUE #60: Fix project list display issues
(Laura first found this - I'm able to replicate occasionally on an actual device but not on simulator)
Project list seems "stuck" at whatever the first discipline is chosen.  There should also be a spinner while the list is loading.

robinschaaf:
Fixes #60 
When quickly navigating around, the router is not always doing component unmounts/mounts (https://github.com/aksonov/react-native-router-flux/issues/415) and so the project list isn't properly refreshing
This PR moves more of the logic for filtering project lists into redux, and fetches once on application mount, then just filters the list displayed once a discipline is chosen (rather than an API call each time the component is mounted)

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -13,8 +13,10 @@ import auth from 'panoptes-client/lib/auth'
 import apiClient from 'panoptes-client/lib/api-client'
 import store from 'react-native-simple-store'
 import { PUBLICATIONS } from '../constants/publications'
+import { MOBILE_PROJECTS } from '../constants/mobile_projects'
+import { GLOBALS } from '../constants/globals'
 import { NetInfo } from 'react-native'
-import { addIndex, forEach, head, keys, map } from 'ramda'
+import { addIndex, filter, forEach, head, keys, map, propEq } from 'ramda'
 import { Actions, ActionConst } from 'react-native-router-flux'
 
 export function setState(stateKey, value) {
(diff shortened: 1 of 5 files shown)


src/actions/index.js
```diff
@@ -117,13 +119,21 @@ export function signOut() {
...
     dispatch(setError(''))
-    dispatch(setIsFetching(false))
+    var callFetchProjects = tag => dispatch(fetchProjectsByTag(tag.value))
+    forEach(callFetchProjects, filter(propEq('display', true), GLOBALS.DISCIPLINES))
+  }
+}
+
+
+export function fetchProjectsByTag(tag) {
+  const parms = {id: MOBILE_PROJECTS, cards: true, tags: tag, sort: 'display_name'}
```
wgranger: What did we decide these mobile projects will be if the environment isn't set to production? I see there's a list of proj IDs, but I'm not clear what they are pointing to on staging.

wgranger: Right. I saw this was pointing to `mobile_projects` file, I was just curious what those project IDs represent when it isn't set to production in that ternary. I guess that's something to think about but in no way blocking for this PR.

robinschaaf: Yeah, I think they actually worked as a random list of projects at one point, but I think when we added the 'tags' query they stopped working altogether :-(  Right now I have the Panoptes client hardcoded to production in the /node_modules/, which is less than desirable, as I have to redo it everytime npm install is run 👎

src/components/ProjectDisciplines.js
```diff
@@ -37,7 +44,8 @@ class ProjectDisciplines extends React.Component {
           title={label}
           tag={value}
           key={idx}
-          color={color} /> )
+          color={color}
+          setSelectedProjectTag={() => {this.props.setSelectedProjectTag(value)}} /> )
```
wgranger: Should this prop `setSelectedProjectTag` be defined in the `Discipline` component? I'm not seeing any reference to it there, but perhaps I'm missing something.

robinschaaf: Hmm, it's in the proptypes and it's being called in handleClick.  Is that what you were looking for or ?

wgranger: You are correct. It's been a long day.

PR #64, opened 2016-11-21
Add jest snapshot unit tests

### LINKED ISSUE #40: Testing, linting etc
- [ ] Add tests with Jest
- [x] Add to Travis for test running
- [x] Add eslint config for react native

robinschaaf:
Add tests using jest's snapshot for components and reducers.  
More info:
https://facebook.github.io/jest/docs/tutorial-react-native.html

A few updates to existing components were also made to make them more easily testable:
  *  Exporting "undecorated" components so that the redux connect to provider isn't required to test
  *  Move logic and redux store from PublicationFilter to PublicationList, to simplify that component and so that it doesn't have to be mocked as a child component with the redux store connection
  *  Move render of nav bar on signin to use the static function like the other components (to make it work the same as other components and remove the need to mock)

(also updates readme)
Fixes #40

### DIFF
--- a/src/components/__tests__/Discipline-test.js
+++ b/src/components/__tests__/Discipline-test.js
@@ -1,0 +1,11 @@
+import 'react-native'
+import React from 'react'
+import renderer from 'react-test-renderer'
+import Discipline from '../Discipline'
+
+it('renders with a title', () => {
+  const tree = renderer.create(
+    <Discipline title='O Hai!' />
+  ).toJSON()
+  expect(tree).toMatchSnapshot()
+})
--- a/src/components/__tests__/Discipline-test.js
+++ b/src/components/__tests__/Discipline-test.js
@@ -5,7 +5,12 @@ import Discipline from '../Discipline'
 
 it('renders with a title', () => {
   const tree = renderer.create(
-    <Discipline title='O Hai!' />
+    <Discipline
+      title='O Hai!'
+      icon='arts'
+      tag='test'
+      color='#fff'
+      setSelectedProjectTag={jest.fn} />
   ).toJSON()
   expect(tree).toMatchSnapshot()
 })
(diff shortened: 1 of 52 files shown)


src/components/__tests__/Discipline-test.js
```diff
@@ -0,0 +1,16 @@
...
+it('renders with a title', () => {
+  const tree = renderer.create(
+    <Discipline
+      title='O Hai!'
+      icon='arts'
+      tag='test'
+      color='#fff'
+      setSelectedProjectTag={jest.fn} />
+  ).toJSON()
+  expect(tree).toMatchSnapshot()
```
wgranger: In a nutshell, these snapshots are testing that the component, `Discipline` in this case, is rendering as expected based on the props passed in. Is that a correct assumption of these snapshots?

robinschaaf: Yes, and then in the future if the rendered output is different, it will cause the test to fail (meaning it needs to be re-generated)

src/reducers/__tests__/reducers-test.js
```diff
@@ -0,0 +1,20 @@
...
+
+it('unsets user', () => {
+  expect(reducers(initialState, setUser({}))).toMatchSnapshot()
+})
+
+it('sets state from initial state', () => {
+  expect(reducers(initialState, setState('isConnected', true))).toMatchSnapshot()
+})
+
+it('sets a new state', () => {
```
wgranger: These match with the snapshots directly? Ex: this matches with what is expected in the object found at line 1 of `reducers-test.js.snap`?

wgranger:
Ok, this looks to be many lines of code for what actually seems pretty straightforward. If I understand correctly, jest will always check against the snapshots created in this PR to make sure components stay the same as we proceed from this point. LGTM.

PR #66, opened 2016-11-22
Add check on app state change to reload user data

### LINKED ISSUE #65: Refresh user data upon app launch
User data (avatars, project data, stats) should be updated upon app launch (either opening after being closed entirely or once it becomes active again after pressing home)

Currently data is only being refreshed when logged out and logged back on.  But since we're not requiring login on app launch this data can get stale.

robinschaaf:
Fixes #65 
When an app state changes to 'active', this will reload user avatar data (eventually this will be used to reload other data, such as project preferences).  This app state listener will fire when users press the home button and then reopens the app later (state changes from background to active)

This PR also refactors code a bit to rely more heavily on promises, which will make adding project preferences and stats easier later.

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -43,9 +43,10 @@ export function setProjectList(projectList) {
   return { type: SET_PROJECT_LIST, projectList }
 }
 
-export function storeUser(user) {
-  return () => {
-    store.save('@zooniverse:user', {
+export function syncUserStore() {
+  return (dispatch, getState) => {
+    const user = getState().user
+    return store.save('@zooniverse:user', {
       user
     })
   }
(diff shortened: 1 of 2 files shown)


src/actions/index.js
```diff
@@ -43,69 +43,118 @@ export function setProjectList(projectList) {
...
-          })
-      } else {
-        dispatch(setError('Sorry, but you must be connected to the internet to use Zooniverse'))
+    dispatch(checkIsConnected()).then(() => {
+      auth.signIn({login: login, password: password}).then((user) => {
+        user.isGuestUser = false
+        dispatch(setUser(user))
+
+        return Promise.all([
+          dispatch(loadUserAvatar()) //will have more added
```
wgranger: What will be added here? It seems a bit awkward to use `Promise.all` with an array of length one.

robinschaaf: Yep, definitely!  This was just to set it up so that future PRs that build off this code will have fewer diffs (I have multiple branches that work off this same code).  It eventually gets the user's projects (for calculating recent projects and classification count) and the project level notification settings.

wgranger: Ahh, ok cool. I suppose at the point of this PR avatar is the only user specific resource set up.

robinschaaf: Yeah, exactly.

wgranger:
Looks good. Things appear to be up to speed on my end with Android Studio as well (after a huge update).

PR #69, opened 2016-11-29
Fix landscape issues - user avatar positioning and status bar disappearing

### LINKED ISSUE #68: Fix landscape view
Avatar isn't re-centering when rotating to landscape.  Also status bar is disappearing on iOS.

screenshot_1480362747

robinschaaf:
Two landscape fixes:
  *  Use flex to center avatar
  *  Status bar is by default hidden in landscape mode in iOS - this adds it back
Fixes #68


wgranger:
This PR looks simple, so I'm going to go ahead and merge it. I'd double check that an iOS device can switch to landscape mode (not affected by this branch/PR). I'll open an issue on the matter.

PR #71, opened 2016-11-30
Add scrolling to sign in

### LINKED ISSUE #70: Add scrolling to sign in screen
For smaller devices, in landscape mode, or when the keyboard is expanded, user is unable to scroll the screen.  In smaller devices the screen is actually cut off.

robinschaaf:
Adds a ScrollView container to the sign in page to allow scrolling in case the screen is too small to show the entire sign in page. 
Also adds a top padding calculation to take into account iOS (with the status bar) or Android.

Fixes #70

### DIFF
--- a/src/components/SignIn.js
+++ b/src/components/SignIn.js
@@ -2,6 +2,8 @@ import React from 'react'
 import {
   Alert,
   Linking,
+  Platform,
+  ScrollView,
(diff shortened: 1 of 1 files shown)


wgranger:
LGTM. ScrollView is magic.

PR #74, opened 2016-12-08
Open project directly from project-specific push notification

### LINKED ISSUE #72: Direct project-specific push notifications to project landing page
For push notifications that are for mobile-friendly projects, we will automatically redirect the push notification to open that project directly when clicked on / swiped.  This will require adding a project ID to the push notification payload.

robinschaaf:
*  Adds support to handle project_id in push notification payload
  *  Uses PushNotificationIOS from react-native for iOS and react-native-fcm for android/firebase
  *  On Android, when app is in foreground, sends a local notification to display in the system tray
  *  When project ID is present and one that is listed in the Mobile-Friendly projects) it will directly open that project in the browser

Fixes #72

### DIFF
--- a/android/app/src/main/java/com/zooniversemobile/MainActivity.java
+++ b/android/app/src/main/java/com/zooniversemobile/MainActivity.java
@@ -1,8 +1,15 @@
 package com.zooniversemobile;
 
+import android.app.Notification;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
 import android.content.Intent;
 import android.content.res.Configuration;
+import android.media.RingtoneManager;
+import android.net.Uri;
 import android.os.Bundle;
+import android.support.v4.app.NotificationCompat;
 
 import com.facebook.react.ReactActivity;
 import com.google.android.gms.common.ConnectionResult;
@@ -34,6 +41,7 @@ public class MainActivity extends ReactActivity {
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+
         if (playServicesAvailable()) {
             PusherAndroid pusher = new PusherAndroid(BuildConfig.PUSHER_API_KEY);
             final PushNotificationRegistration nativePusher = pusher.nativePusher();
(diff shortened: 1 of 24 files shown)


src/components/StyledText.js
```diff
@@ -3,9 +3,11 @@ import {
   Text
 } from 'react-native'
 import EStyleSheet from 'react-native-extended-stylesheet'
+import { append } from 'ramda'
 
 const StyledText = (props) => {
-  const textStyle = ( props.textStyle ? [styles.defaultText, styles[props.textStyle]] : styles.defaultText )
+  var textStyle = ( props.textStyle ? [styles.defaultText, styles[props.textStyle]] : [styles.defaultText] )
```
wgranger: Avoid using `var` here. Use `let` when declaring a variable that will later be reassigned, per ES6 standards (unless RN restricts you from doing it for some reason).

src/components/StyledModal.js
```diff
@@ -0,0 +1,72 @@
...
+import Icon from 'react-native-vector-icons/FontAwesome'
+
+class StyledModal extends Component {
+  render() {
+    return (
+      <View>
+        <Modal
+          animationType={'fade'}
+          transparent={true}
+          onRequestClose={() => {}}
```
wgranger: What is the purpose of assigning an empty function? Is this a required prop?

src/components/NotificationModal.js
```diff
@@ -0,0 +1,144 @@
...
+import StyledText from './StyledText'
+import Button from './Button'
+import { connect } from 'react-redux'
+import { fetchProject } from '../actions/index'
+import { MOBILE_PROJECTS } from '../constants/mobile_projects'
+import { indexOf, isEmpty } from 'ramda'
+import GoogleAnalytics from 'react-native-google-analytics-bridge'
+
+
+var notificationTitle, notificationBody, projectID
```
wgranger: Use `let` if you're able.

android/app/src/main/java/com/zooniversemobile/MainActivity.java
```diff
@@ -87,6 +97,29 @@ private boolean playServicesAvailable() {
         return true;
     }
 
-
-
+    private void sendNotification(String messageTitle, String messageBody, String projectID) {
+        Intent intent = new Intent(this, MainActivity.class);
```
wgranger: Maybe choose a different name for your Intent here? `Intent intent` just seems a bit odd semantically.

wgranger:
The code looks good despite a couple questions about syntax, etc. Is there any simple way to test or trigger a notification on this branch?

src/components/StyledText.js
robinschaaf: Nope - just a habit I need to break!  I'll update these

src/components/StyledModal.js
robinschaaf: Yes, on android, according to the documentation:
http://facebook.github.io/react-native/releases/0.37/docs/modal.html

robinschaaf:
Unfortunately there isn't a simple way to test, now that the push certificate is for production (and pusher only has one option for which certificate to use).  the only way I can think of right now is for us to subscribe to a pusher interest that's for dev only.  I'll open a support ticket with pusher to see how they recommend testing in dev when there's a production instance running.

src/components/NotificationModal.js
```diff
@@ -0,0 +1,144 @@
...
+        <View style={styles.container}>
+          <StyledText
+            text={notificationBody} />
+        </View>
+
+        { (isEmpty(this.props.notificationProject) || (isEmpty(projectID))) ? null : projectDisplay }
+
+        <Button
+          handlePress={() => this.props.setVisibility(false)}
+          buttonStyle={'greyButton'}
```
wgranger: Final note: change this `buttonStyle` because now it looks like a disabled button.

robinschaaf: Just updated it.  Thanks!

PR #76, opened 2017-01-31
Split actions/index into user and auth

robinschaaf:
actions/index.js has been getting large (and will just be getting larger) so this PR starts separating them out into actions/auth.js and actions/user.js. actions/index.js will be used for global action creators (such as retrieving the Publications list).


src/actions/auth.js
```diff
@@ -0,0 +1,62 @@
+import auth from 'panoptes-client/lib/auth'
+import store from 'react-native-simple-store'
+import { Actions, ActionConst } from 'react-native-router-flux'
+
+import { checkIsConnected, setState, setIsFetching } from '../actions/index'
+import { loadUserAvatar, syncUserStore } from '../actions/user'
+
+export function getAuthUser() {
+  return () => {
```
wgranger: Is there a reason this promise is written inside this extra function? I'm sure this is purposeful, I'm mainly curious what the reason is.

wgranger:
Good thinking. LGTM, but let me know if I need to wait for the other PR to be addressed before merging this.

robinschaaf:
Good point, it'll cause merge conflicts either way.  If you want to just merge this one first then I'll rebase and fix the other comments from the other one.

PR #77, opened 2017-02-01
Add Checkbox Component and Registration

robinschaaf:
Adds Register component
Adds Checkbox component
Adds functionality to Input component to accept a label/sub-label, new keyboard style
Add new state for Registration to capture data entry
Add/update theme colors for the Registration button

Adds new tests for Register and Checkbox, and new utils for checking valid email and valid login
Updates snapshots for tests that changed due to updated Input component (SignIn and Input) and reducers test for new Registration initial state

### DIFF
--- a/src/components/Input.js
+++ b/src/components/Input.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import {
+  StyleSheet,
   TextInput,
   View
 } from 'react-native'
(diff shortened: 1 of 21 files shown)


src/components/Input.js
```diff
@@ -1,47 +1,94 @@
...
-  passwordField: React.PropTypes.bool
+  passwordField: React.PropTypes.bool,
+  keyboardType: React.PropTypes.string,
+  inputStyle: React.PropTypes.string,
+  subLabelText: React.PropTypes.string,
+  addLabel: React.PropTypes.string
 }
 
-Input.defaultProps = { labelText: '', passwordField: false };
+Input.defaultProps = { labelText: '', passwordField: false, required: false };
```
wgranger: This is me being a bit pedantic, but maybe keep the style of `defaultProps` as with `propTypes` above (putting one key/value of an object per line. I don't think it's necessarily worth a separate commit for just that purpose, but just for the future.

robinschaaf: Ahh, yeah I see

src/components/Register.js
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+
+  static renderNavigationBar() {
+    return <NavBar showDrawer={false} showBack={true} title={'Register'} />;
+  }
+
+  render() {
+    const registerDisabled = any(converge(or, [isEmpty, isNil]), [this.props.registration.login, this.props.registration.password, this.props.registration.email])
+    const emailChecked = this.props.registration.global_email_communication
```
wgranger: Might be good to give a default prop to `registration`. I can see this throwing an error when trying to look for the `global_email_communication` tag if there isn't a registration prop present.

robinschaaf: It has the default value set up in the reducer initial state (you can see in reducers/index.js)

wgranger:
Looks good. Just one small line comment. I like the change in color of the registration button. Also, good use of regex for validations.

src/components/Register.js
wgranger: Ok. Yes, thought it might be the case that I was missing something with Redux. Thanks.

PR #78, opened 2017-02-02
Updates to allow for dev version of Push notifications

### LINKED ISSUE #79: Need to be able to send push notifications in dev
Now that the app is in production we need to be able to send separate push notifications for dev-only.  According to Pusher, the easiest way to do this is to add a new pusher "app" for dev, with new  keys, to send to.  I've added the new app in pusher (called "zooniverse-mobile-dev").  In order to send separately in Apple we need to use the dev certificate, and in order to get new FCM keys (for android) I had to set up a new Firebase project, called "ZooniverseMobileDev".  We then need to update the application to use the new dev version of google-services.json and update the Pusher API Key to use a different one when running in "debug" mode.

robinschaaf:
In order to have a separate API key for push notifications in both dev and prod, we need to have a separate firebase project for development only, which is set up by google-services.json config files.  Then we can have the release version kept separately from the dev version, per the following document:  https://developers.google.com/android/guides/google-services-plugin

This PR moves the current google-services.json from app/ into app/src/release/ and adds a new google-services.json for the new ZooniverseMobileDev Firebase project.  It also adds a new PUSHER_API_KEY for dev. (You'll need to add a PUSHER_API_KEY_DEV to ~/.gradle/gradle.properties - I can give you the value).  It also adds the new Pusher API Key for dev on iOS.

Fixes #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ You'll also need at least one emulator.  I've found the easiest way to add Virtu
 Also you'll need a Gradle properties file outside the project for keeping secrets that aren't checked into source control.  This should be in ~/.gradle/gradle.properties and contain the following:
 `MYAPP_RELEASE_STORE_FILE=
 MYAPP_RELEASE_KEY_ALIAS=
-PUSHER_API_KEY=`
+PUSHER_API_KEY=
+PUSHER_API_KEY_DEV=`
 
 ## Run
 npm install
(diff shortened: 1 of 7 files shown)


wgranger:
I would make a note about this in the README, because this won't build in Android Studio without the `PUSHER_API_KEY_DEV`.

robinschaaf:
Good catch! Updated.

wgranger:
Ok. With the server running, this seems fine on Android.

PR #80, opened 2017-02-06
Add Circle Ribbon SVG

### LINKED ISSUE #36: Add user avatar stats circle
Add the colored circle around the user avatar like in the mockup.  Can reference PFE code for it:

iphone 6 copy 34

robinschaaf:
Add react-native-svg package from node for SVG support (RN doesn't have this built in)
Load user project data for use with the Circle Ribbon (user project data will also be used for other pieces)
Add CircleRibbon component to display the ribbon around the user avatar, and update some styles as a result
Add react-native-svg mocks and test for CircleRibbon
Fixes #36

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "react-native-orientation": "^1.17.0",
     "react-native-router-flux": "^3.35.0",
     "react-native-simple-store": "^1.1.0",
-    "react-native-svg": "^4.3.3",
+    "react-native-svg": "~4.3.3",
     "react-native-vector-icons": "^2.1.0",
     "react-native-webview-bridge": "^0.33.0",
     "react-redux": "^4.4.5",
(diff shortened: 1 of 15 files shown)


wgranger:
I'm seeing some of these snapshots fail in testing. Might be worth a double check.

robinschaaf:
Whoops, forgot to update them.  Should be good now.  Good catch!

package.json
```diff
@@ -24,6 +24,7 @@
     "react-native-orientation": "^1.17.0",
     "react-native-router-flux": "^3.35.0",
     "react-native-simple-store": "^1.1.0",
+    "react-native-svg": "~4.3.3",
```
wgranger: **Note**: this initially caused some dependency issues when clearing node_modules and installing again (fixed with later commit). We should add a checklist for PRs to ensure things work on each branch after a new `npm install`.

PR #82, opened 2017-02-08
Add webview view for PFE projects

robinschaaf:
Adds a Webview component for viewing PFE projects within Webview instead of going outside the app to the default browser.  For non-PFE projects it still opens in the browser.

Since page loads within PFE are mostly from ajax requests, natively they're not processed as new page hits, and thus the router wasn't capturing them all to populate the action stack.  This made the back button not work when navigating between pages.  This also adds logic for custom back handling by adding event listeners to "a" links that fired a message over a bridge to notify the app that a link has been clicked.

### DIFF
--- a/src/components/Project.js
+++ b/src/components/Project.js
@@ -9,6 +9,7 @@ import {
 } from 'react-native';
 import EStyleSheet from 'react-native-extended-stylesheet'
 import GoogleAnalytics from 'react-native-google-analytics-bridge'
+import {Actions} from 'react-native-router-flux'
 import Icon from 'react-native-vector-icons/FontAwesome'
 
 class Project extends Component {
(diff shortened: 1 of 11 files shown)


src/components/Project.js
```diff
@@ -19,11 +20,18 @@ class Project extends Component {
 
   handleClick() {
     GoogleAnalytics.trackEvent('view', this.props.project.display_name)
+    if (this.props.project.redirect) {
```
wgranger: Look into using `.shape` when setting propTypes for items like objects. It's a way to further define ambiguous props like objects, so you could then tell it to expect a `redirect` tag. We've recently started doing this in PFE. This is probably fine for now.

wgranger:
This code looks fine and most of it seems to be setup. It's a bit difficult to test now without the production projects available.

PR #84, opened 2017-02-10
Add PR Template

robinschaaf:
Let me know if you think there should be any other checks!


wgranger:
Perfect. This looks good for now. We can adjust them in the future if we need.

PR #85, opened 2017-02-13
Styling updates

### LINKED ISSUE #52: Jordan's styling feedback
*  The project card should be updated to include description (see attached), and the color of the project card arrow should coincide with the color for that discipline (rather than just the green all the time)
  *  The zoo icons should be larger
  *  The "Registration" button on the signin page should use dark blue rgba(12, 72, 129, 1) instead of the red
  *  User avatar should be updated to the "ninja" (attached below)
  *  Guest user should have a zooniverse logo rather than avatar (not received)

robinschaaf:
Fixes #52 

Describe your changes.
Includes styling changes from feedback from Jordan listed in Issue

Also some general styling/ux updates
  *  Add text on loading screen about where it is in the loading process (good for users with a lot of data, it can take a while to load everything when logging in)
  *  Fade in project cards once loaded so it's not so jarring when they appear
  *  Make entire project card tappable, rather than just the title

# Review Checklist

- [x] Are tests passing?


src/components/Discipline.js
```diff
@@ -57,17 +57,17 @@ const styles = EStyleSheet.create({
     fontFamily: 'OpenSans-Semibold',
     backgroundColor: '$transparent',
     color: '$textColor',
-    fontSize: 22,
-    lineHeight: 30,
-    width: '100% - 120'
+    fontSize: 28,
+    lineHeight: 38,
+    width: '100% - 140'
```
wgranger: I'm sure you've already checked this, but are you sure this calculation is working correctly? Just seems a bit odd to have `'100% - 140'`, semantically. Is this to adjust the size of items responsively depending on screen size?

src/components/Project.js
```diff
@@ -44,46 +60,67 @@ class Project extends Component {
   render() {
     const imageURI = `https://${this.props.project.avatar_src}`
```
wgranger: It would be good to have a default image here if there isn't one for `project.avatar_src`. The teal_wallpaper.png image above would probably work.

src/components/Discipline.js
robinschaaf: Yeah, I think this looks weird in RN since they don't use units, plus percents only work on the full device size, not container size.  Subtracting 140 is to get rid of the width of the Zoo icon container (80), the right arrow icon (20) and the right / left margins and padding (50).  Although this adds up to 150 now that I'm looking at it. 👎   I'll fix that and also convert these values into variables, I think having them named may help this calculated value make more sense and be less math-error-prone.

src/components/Project.js
robinschaaf: Good idea!

wgranger:
Looks good

PR #87, opened 2017-02-13
Add onboarding screen

### LINKED ISSUE #86: Add onboarding screen
Chris suggested adding an onboarding screen to introduce new users to Zooniverse  - in case they downloaded the app for any reason and aren't sure or don't remember what it is when they open it up.

robinschaaf:
Fixes #86 

Add an onboarding screen for new users.  Includes the following changes:
  *  Add images from Jordan to randomly display a different image
  *  Update the timing and add fading of the launch screen on ios to give react app enough time to initialize
  *  Add LaunchScreen component for both iOS and Android to avoid flashing white while loading Onboarding screen
  *  Since we don't want the status bar displayed on the onboarding screen I had to take it off the global index.ios.js, and all screens need to use renderNavigationBar, so ProjectDiscipline is updated to use this method

# Review Checklist

- [x] Are tests passing?


src/components/Button.js
```diff
@@ -4,17 +4,21 @@ import {
   TouchableOpacity
 } from 'react-native'
 import EStyleSheet from 'react-native-extended-stylesheet'
+import { append } from 'ramda'
 
 const Button = (props) => {
-  const buttonStyle = ( props.buttonStyle ? [styles.button, styles[props.buttonStyle]] : styles.button )
+  var buttonStyle = ( props.buttonStyle ? [styles.button, styles[props.buttonStyle]] : styles.button )
```
wgranger: use `let` rather than `var`

src/components/Button.js
```diff
@@ -4,17 +4,21 @@ import {
...
+  buttonStyle = (props.additionalStyles ? append(props.additionalStyles, [buttonStyle]) : buttonStyle)
+
+  var textStyle = (props.additionalTextStyles ? append(props.additionalTextStyles, [styles.buttonText]) : styles.buttonText)
```
wgranger: You should be able to use `const` here

src/components/Launch.js
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <Image source={require('../../images/fulllaunch.png')} style={styles.container} />
+  )
+}
+
+const styles = EStyleSheet.create({
+  container: {
+    flex: 1,
+    width: null,
+    height: null,
```
wgranger: This seems kind of odd. Is this the standard way to have an image take up the entire page rather than using 100%?

src/components/Button.js
robinschaaf: Whoops good catch!

src/components/Launch.js
robinschaaf: Yeah it's an RN thing :-( https://github.com/facebook/react-native/issues/950 Basically it resets any sizing and forces RN to calculate the image size based on device size (alternatively I could pull in the device size and manually calculate the size of the image myself).  Otherwise trying to position: absolute doesn't work, neither does just having flex: 1 and omitting the width/height.

wgranger: Nah, I think this is fine. It seems a bit off from working on PFE so much, but this makes sense if it's particular to RN.

wgranger:
Looks good.

PR #90, opened 2017-02-14
Add notifications settings screen

robinschaaf:
Adds the Notification Settings screen (on the side menu) to allow users to subscribe or unsubscribe to General Zooniverse Notifications or specific projects (populated from the list of Mobile-Friendly Projects).
Note that the notification settings are saved at the device level, since the list of projects are based on the Mobile friendly projects (and not the user preferences), and also notifications subscriptions are stored at the device level with Pusher.

  *  Add support in android and ios for updating interest subscriptions in pusher (pusher doesn't have a react-native library, only native libraries)
  *  Use NativeModules (RN) to send message from RN to Native to update the Pusher "interest" subscriptions when checked/unchecked
  *  For iOS, if push isn't enabled, display a message explaining how to enable them
  *  Load notification settings upon signing in, or refresh
  *  Store notification settings in local storage, separate from user
  *  Add/Update tests

This can be tested using the send-push.js script, changing the to_notify to the project ID you wish to test.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/NotificationSettings.js
+++ b/src/components/NotificationSettings.js
@@ -51,12 +51,12 @@ export class NotificationSettings extends React.Component {
 
 
   render() {
-    var mobileProjects = flatten(
+    let mobileProjects = flatten(
       map((tag) => { return this.props.projectList[tag] }, keys(this.props.projectList))
     )
 
     const renderPreference = (id, idx) => {
-      var project = find(propEq('id', id))(mobileProjects)
+      let project = find(propEq('id', id))(mobileProjects)
 
       if (project === undefined) { //project may no longer exist
         return
@@ -87,7 +87,7 @@ export class NotificationSettings extends React.Component {
         <View style={styles.switchContainer}>
           <Switch
             value={this.props.notifications['general']}
-            style={styles.switch}
+            style={styles.switchComponent}
             onTintColor={theme.headerColor}
             onValueChange={(checked) => this.props.updateGeneralNotification(checked)}
           />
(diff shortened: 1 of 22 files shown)


src/components/ProjectNotification.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+const styles = EStyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    padding: 8,
+    alignItems: 'center'
+  },
+  switch: {
```
wgranger: `switch` is a reserved word in JS, so might be better to choose something different.

src/components/NotificationSettings.js
```diff
@@ -0,0 +1,153 @@
...
+    this.props.checkPushPermissions()
+  }
+
+  static renderNavigationBar() {
+    return <NavBar title={'Notification Settings'} showBack={true} />;
+  }
+
+
+  render() {
+    var mobileProjects = flatten(
```
wgranger: `let`

src/components/NotificationSettings.js
```diff
@@ -0,0 +1,153 @@
...
+      map((tag) => { return this.props.projectList[tag] }, keys(this.props.projectList))
+    )
+
+    const renderPreference = (id, idx) => {
+      var project = find(propEq('id', id))(mobileProjects)
```
wgranger: `let`

src/components/NotificationSettings.js
```diff
@@ -0,0 +1,153 @@
...
+  switchContainer: {
+    flexDirection: 'row',
+    paddingLeft: 8,
+    paddingBottom: 16,
+    paddingTop: 16,
+    alignItems: 'center',
+    borderBottomColor: '$lightGrey',
+    borderBottomWidth: StyleSheet.hairlineWidth,
+  },
+  switch: {
```
wgranger: Again, careful, as this is a reserved word.

wgranger:
All set.

PR #93, opened 2017-02-16
Add Permissions Primer

robinschaaf:
Since once iOS prompts users for push notifications only once, and by default on application load, it can be a better experience and less jarring to move the permissions until later, and also to explain what notifications will be used for before prompting.
This PR:
  *  Moves to use RN's PushNotificationIOS so that the timing of the prompt is controlled on the react-native side
  *  Also uses RN alert to give the explanation and store within redux whether the user has been prompted, so it doesn't re-prompt.  (As it works currently we can't re-ask a user to enable notifications - but this way we can if they decide later they would like to enable them)

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/ProjectDisciplines.js
+++ b/src/components/ProjectDisciplines.js
@@ -67,7 +67,7 @@ export class ProjectDisciplines extends React.Component {
           'Zooniverse would like to occasionally send you info about new projects or projects needing help.',
           [
             {text: 'Not Now', onPress: () => this.requestIOSPermissions(false)},
-            {text: 'Sure!', onPress: () => this.requestIOSPermissions(true)},
+            {text: 'OK', onPress: () => this.requestIOSPermissions(true)},
           ]
         )
       }
--- a/src/components/ProjectDisciplines.js
+++ b/src/components/ProjectDisciplines.js
@@ -1,6 +1,8 @@
 import React from 'react'
 import {
+  AlertIOS,
   Platform,
+  PushNotificationIOS,
   ScrollView,
   StyleSheet,
   View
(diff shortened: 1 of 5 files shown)


src/components/ProjectDisciplines.js
```diff
@@ -25,20 +28,59 @@ const mapStateToProps = (state) => ({
...
+
+  promptRequestPermissions = () => {
+    PushNotificationIOS.checkPermissions((permissions) => {
+      if (permissions.alert === 0){
+        AlertIOS.alert(
+          'Allow Notifications?',
+          'Zooniverse would like to occasionally send you info about new projects or projects needing help.',
+          [
+            {text: 'Not Now', onPress: () => this.requestIOSPermissions(false)},
+            {text: 'Sure!', onPress: () => this.requestIOSPermissions(true)},
```
wgranger: `Sure!` seems a bit awkward here. Can we make it `Ok`? Seems to be a bit more standard.

wgranger:
When I reinstall the app on my device `iOS` I get two prompts upon the initial sign in. One asks: "Allow Notifications: Zooniverse would like to occasionally send you info about new projects or projects needing help." After that, another one reads, "'Zooniverse' would like to send you notifications: Notifications may include alerts, sounds, and icon badges. These can be configured in Settings." 

Is this the expected behavior?

robinschaaf:
Yes, the first is the custom one to notify the user that the permission prompt is coming, and the second one is the standard iOS prompt.  Just to verify, it happened after you signed in?  (this is the expected behavior)

src/components/ProjectDisciplines.js
robinschaaf: Sure

wgranger:
Yes, it happened right after I signed in. Alright, just making sure. Everything else on this PR looked alright.

PR #94, opened 2017-02-16
Changes to offline handling

### LINKED ISSUE #42: Improve UX on offline notification warning
Just shows some text for now, but will affect things like notifications settings etc

robinschaaf:
Fixes #42 

Describe your changes.
  *  Remove the in-component Offline messaging to instead just display an error message when the user goes offline (this seems more consistent with how other apps handle offline messaging - such as CNN, Target, etc. and also simplifies component code)
  *  Store project list in local storage to fall back on in case of slow or no internet connection, rather than making it vanish altogether
  *  Default the avatar image source in case the user is offline and we can't download the avatar image from the server
  *  Update tests

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/UserAvatar.js
+++ b/src/components/UserAvatar.js
@@ -7,15 +7,17 @@ import EStyleSheet from 'react-native-extended-stylesheet'
 
 class UserAvatar extends Component {
   render() {
-
     const guestUserLogo =
       <Image source={require('../../images/simple-avatar.png')} style={styles.avatar} />
 
-    const avatar = ( <Image style={styles.avatar} source={
-        this.props.avatar
-        ? {uri: this.props.avatar}
-        : require('../../images/simple-avatar.png') } /> )
-
+    const avatar = (
+      <Image
+        defaultSource={require('../../images/simple-avatar.png')}
+        style={styles.avatar}
+        source={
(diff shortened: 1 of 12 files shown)


src/components/UserAvatar.js
```diff
@@ -7,15 +7,17 @@ import EStyleSheet from 'react-native-extended-stylesheet'
...
       <Image source={require('../../images/simple-avatar.png')} style={styles.avatar} />
 
-    const avatar = ( <Image style={styles.avatar} source={
-        this.props.avatar
-        ? {uri: this.props.avatar}
-        : require('../../images/simple-avatar.png') } /> )
-
+    const avatar = (
+      <Image
+        defaultSource={require('../../images/simple-avatar.png')}
```
wgranger: Why is `defaultSource` needed if there is a ternary to find the same image on `source`?

robinschaaf: That's in case there is a source that exists (the src is stored in local storage) so it passes the ternary, but we're unable to actually download it because of slow or no internet.  This is an RN helper with Image.

wgranger: Gotcha.

PR #100, opened 2017-03-20
Add crypto-js, session generator

robinschaaf:
Describe your changes.
Implements session generation for storing with classification data in the same manor as PFE.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/actions/session.js
+++ b/src/actions/session.js
@@ -16,7 +16,7 @@ export function setSession() {
     store.get('@zooniverse:session').then(json => {
       stored = json.session
       if (stored.ttl < Date.now()) {
-        stored = generateSessionID()
+        stored.id = generateSessionID().id
       } else {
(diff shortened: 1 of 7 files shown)


src/actions/session.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+export function setSession() {
+  return dispatch => {
+    let stored = {}
+    store.get('@zooniverse:session').then(json => {
+      stored = json.session
+      if (stored.ttl < Date.now()) {
+        stored = generateSessionID()
```
wgranger: Should the `ttl` get reset here as well? It looks like on pfe we're only having the sessionID reset at this point, not creating an entirely new object.

robinschaaf: Yes - that's correct.  I've just fixed it.

PR #101, opened 2017-03-21
Add workflow prompt and mobile friendly icon

robinschaaf:
Describe your changes.
  *  Loads project workflows for projects that have mobile workflows (to be used for the prompt)
  *  Adds checks on Project to determine whether or not workflows should be prompted for in a modal (if there are mobile and non-mobile workflows and prompt for workflow setting is true)
  *  Adds workflow prompt modal
  *  Display links to mobile workflows on the project card when there's more than one
  *  If there's more than one mobile workflow, display an error message if the user doesn't click the link (tries to click the image instead)

Note: I could split this up into two or three PRs if you'd prefer - just let me know

# Review Checklist

- [x] Are tests passing?


src/actions/index.js
```diff
@@ -13,9 +13,10 @@ import store from 'react-native-simple-store'
 import apiClient from 'panoptes-client/lib/api-client'
 import { PUBLICATIONS } from '../constants/publications'
 import { MOBILE_PROJECTS } from '../constants/mobile_projects'
+import { SWIPE_WORKFLOWS } from '../constants/mobile_projects'
```
wgranger: You should be able to combine lines 15-16 since they're drawing from the same location.

robinschaaf: Right!  Thanks.

src/actions/index.js
```diff
@@ -142,6 +143,34 @@ export function fetchProjectsByParms(tag) {
...
+      forEach(getWorkflows, swipeProjects)
+      return resolve()
+    })
+  }
+}
+
+export function fetchProjectWorkflows(project) {
+  return dispatch => {
+    return new Promise((resolve) => {
+      apiClient.type('projects').get({id: project.id}).then((projects) => {
```
wgranger: I believe you can also put 'projects' in brackets to avoid the next line using `head(projects)`. 

`.then(([projects]))` -> would give you only item in that array. Just for future reference.

robinschaaf: Oh, yay!  Good to know.

src/components/Project.js
```diff
@@ -33,15 +46,41 @@ class Project extends Component {
...
+  handlePress() {
     GoogleAnalytics.trackEvent('view', this.props.project.display_name)
+
+    const hasSingleMobileWorkflow = length(this.props.mobileWorkflows) === 1
+    const hasMixedWorkflows = length(this.props.mobileWorkflows) > 0 && length(this.props.nonMobileWorkflows) > 0
+
+    if ((hasMixedWorkflows && this.props.promptForWorkflow)) {
+      this.setState({showWorkflowPrompt: true})
+    } else if (hasSingleMobileWorkflow) {
+      //Actions.Classify({ workflowID: head(this.props.mobileWorkflows).id })
```
wgranger: Should this be uncommented? What happens if this condition is met?

robinschaaf: That was just to indicate when the swipe classifier will be started (which is in a different PR).  I can remove it, it will work once that PR  is in.

wgranger: I think Planet Four: Ridges is the only project currently on the app that falls under this condition, and when I click the project, nothing happens. I assume if this condition (`else if (hasSingleMobileWorkflow)`) is removed the workflow will open with `this.openExternalProject()`, so that might be better for now until that swipe PR is merged.

robinschaaf: That makes sense.  It should be for Steller Watch too.  I'll update it though so for now it still works.

robinschaaf: Got a little sidetracked with the other issues, but this is fixed now.

wgranger:
Looks good.

Issue #98, opened 2017-03-20
Clear Classification Counts Upon Signout


wgranger:
<img width="157" alt="screen shot 2017-03-20 at 11 48 29 am" src="https://cloud.githubusercontent.com/assets/14099077/24111129/6fa4b1e0-0d63-11e7-9889-1be48e3e26a2.png">

I'm seeing a landing page that looks like this when I have logged out. This might have to do with some info being saved in local storage and not being cleared when the user has signed out. This is worth replication (sign out, continue as guest, and check landing page), because I have only checked this on the iOS emulator.

robinschaaf:
Somehow I haven't been able to replicate this, although I have seen it in the past.  Something else may have fixed it.  If you see it again, could you try to see if you can replicate it?
Thanks!

wgranger:
I will, but I believe this was a sessions issue that was fixed with #103

PR #103, opened 2017-03-24
Fix issues with session

robinschaaf:
Describe your changes.
  *  Session wasn't being stored when newly generated
  *  Session wasn't being initialized in the reducers
  *  Interpolation was incorrect

# Review Checklist

- [x] Are tests passing?


wgranger:
Fixes #98

PR #105, opened 2017-03-29
Fix at older version of panoptes-client

robinschaaf:
Recently panoptes-client added using window.location:
https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/config.js#L63
which is causing an issue in mobile due to there being no window object since it's not a browser.  This fixes the version at 2.5.2 until a fix is in place.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,8 @@
     "preset": "jest-react-native"
   },
   "dependencies": {
-    "crypto-js": "^3.1.9-1",    
-    "panoptes-client": "^2.5.2",
+    "crypto-js": "^3.1.9-1",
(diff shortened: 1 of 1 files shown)


srallen:
We should remember to ping you for PRs on the JS client to make sure we're not breaking the mobile app.

eatyourgreens:
Fixed in [Panoptes Client 2.7.1](https://github.com/zooniverse/panoptes-javascript-client/commit/2cdddbfb01374860871f8744f640166e3ebbf0ee)

PR #107, opened 2017-04-05
Fix push issues in ios and android

### LINKED ISSUE #67: Android crashing on first install
FCM Listener, onMessageReceived appears to be firing:

                                                                   --------- beginning of crash
11-28 11:14:03.489 2600-3429/com.zooniversemobile E/AndroidRuntime: FATAL EXCEPTION: pool-9-thread-1
                                                                    Process: com.zooniversemobile, PID: 2600
                                                                    java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.google.firebase.messaging.RemoteMessage$Notification.getTitle()' on a null object reference

robinschaaf:
Fixes a few issues in Android (identified by Laura) and that I noticed on iOS while testing that fix
  *  Fix so that Android doesn't send a notification when onMessageReceived unless it has a Notification.  I believe this is happening because pusher is sending a registration success message.  Currently this crashes android upon initial install. [Fixes #67]
  *  Fix so that Android and iOS < 10 doesn't erroneously display an empty message on registration success
  *  Fix iOS so that the interests are synced upon push notification registration.  This is to avoid a  race condition where the interests are synced prior to being prompted for registration, so that if a push notification is sent at the first install (and before any re-syncing occurs) the push notification won't go through.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/containers/zooniverseApp.js
+++ b/src/containers/zooniverseApp.js
@@ -10,8 +10,8 @@ import NotificationModal from '../components/NotificationModal'
 import LaunchScreen from '../components/Launch'
 import NavBar from '../components/NavBar'
 import { connect } from 'react-redux'
-import { setState } from '../actions/index'
-import { isEmpty } from 'ramda'
+import { setState, syncInterestSubscriptions } from '../actions/index'
+import { isEmpty, pathOr } from 'ramda'
 import FCM from 'react-native-fcm'
 
 const mapStateToProps = (state) => ({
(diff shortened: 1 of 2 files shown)


src/containers/zooniverseApp.js
```diff
@@ -39,18 +42,30 @@ class ZooniverseApp extends Component {
...
-    PushNotificationIOS.removeEventListener('notification', this.onRemoteNotificationIOS);
+    PushNotificationIOS.removeEventListener('notification', this.onRemoteNotification);
+    PushNotificationIOS.removeEventListener('register', this.onPushRegistration)
   }
 
   onRemoteNotification = (notification) => {
-    this.props.setNotificationPayload(notification)
-    this.props.setModalVisibility(true)
+    //this is called on iOS < 10 when registered, so make sure it's a valid push notification
+    var isValidIOS = Platform.OS === 'ios' && !pathOr(false, ['_data', 'pusher_token_validation'], notification)
```
wgranger: use `let` here and below.

PR #110, opened 2017-04-11
Store workflow data

robinschaaf:
Project data is stored locally (in case a user has a slow / no internet connection), this adds functionality to store the workflow data as well, which is used to indicate that a project is mobile friendly and displays the mobile icon.  This way in case the api sync hasn't completed before the user visits a category it won't be jumpy or incorrectly go to the PFE project.
(it's probably easiest for comparison to merge #109 first, but I had to incorporate that to get it running in iOS)

# Review Checklist

- [x] Are tests passing?


wgranger:
Looks good. Thanks for the rebase.

PR #111, opened 2017-04-11
Add Classification Panel for Swipe

robinschaaf:
In an attempt to make this PR smaller I've pulled out the presentational base of the swipe app.  This doesn't include the Tutorial, Guide or subject swipe (those will be in later PRs)

Adds markdown support which is used in multiple places in the classifier (the question, field guide, tutorial).   I attempted several react-native markdown libraries but none were close enough or customizable enough to the markdownz rendering - and so I settled for using a combination of webview and the markdown-it library

  *  default-md-html.js used as a basic html template for all markdown being rendered using webview.  It includes a method to set the height because webview requires an explicit height, but we don't know the height until the markdown has been rendered.
  *  remove-md-tab.js is a plugin for markdown-it, copied mostly from the markdownz lib/lib/links-in-new-tabs.js except that it removes the +tab+ since we don't want it opening a new tab for webview

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/actions/classifier.js
+++ b/src/actions/classifier.js
@@ -1,31 +1,33 @@
 import apiClient from 'panoptes-client/lib/api-client'
 import { isNil } from 'ramda'
-import { displayError, setState } from '../actions/index'
+import { setState } from '../actions/index'
 import { Actions } from 'react-native-router-flux'
+import { Alert } from 'react-native'
 
 export function startNewClassification(workflowID) {
   return (dispatch) => {
-    dispatch(setState('classifier.isFetching', true))
     dispatch(setState('loadingText', 'Loading Workflow...'))
     dispatch(setState('classifier.currentWorkflowID', workflowID))
     dispatch(fetchWorkflow(workflowID)).then(() => {
       dispatch(setState('classifier.isFetching', false))
-    }).catch((e) => {
-      dispatch(displayError(`Sorry, but there was an error loading this workflow.  Please try again later. ${e}`))
-      dispatch(setState('classifier.isFetching', false))
-      Actions.pop()  //go back to previous screen
+    }).catch(() => {
+      Alert.alert('Error', 'Sorry, but there was an error loading this workflow.  Please try again later.',
+        [{text: 'Go Back', onPress: () => { Actions.pop()}}]
+      )
     })
   }
 }
 
 export function fetchWorkflow(workflowID) {
   return (dispatch, getState) => {
-    return new Promise ((resolve) => {
+    return new Promise ((resolve, reject) => {
       if (isNil(getState().classifier.workflow[workflowID])) {
         return apiClient.type('workflows').get({id: workflowID}).then(([workflow]) => {
           dispatch(setState(`classifier.workflow.${workflowID}`, workflow))
           dispatch(setState(`classifier.tasks.${workflowID}`, workflow.tasks))
           return resolve()
+        }).catch((e) => {
+          return reject(e)
         })
       } else {
         return resolve()
--- a/src/actions/classifier.js
+++ b/src/actions/classifier.js
@@ -1,0 +1,35 @@
+import apiClient from 'panoptes-client/lib/api-client'
+import { isNil } from 'ramda'
+import { displayError, setState } from '../actions/index'
+import { Actions } from 'react-native-router-flux'
+
+export function startNewClassification(workflowID) {
+  return (dispatch) => {
+    dispatch(setState('classifier.isFetching', true))
+    dispatch(setState('loadingText', 'Loading Workflow...'))
+    dispatch(setState('classifier.currentWorkflowID', workflowID))
+    dispatch(fetchWorkflow(workflowID)).then(() => {
+      dispatch(setState('classifier.isFetching', false))
+    }).catch((e) => {
+      dispatch(displayError(`Sorry, but there was an error loading this workflow.  Please try again later. ${e}`))
+      dispatch(setState('classifier.isFetching', false))
+      Actions.pop()  //go back to previous screen
+    })
+  }
+}
+
+export function fetchWorkflow(workflowID) {
+  return (dispatch, getState) => {
+    return new Promise ((resolve) => {
+      if (isNil(getState().classifier.workflow[workflowID])) {
+        return apiClient.type('workflows').get({id: workflowID}).then(([workflow]) => {
+          dispatch(setState(`classifier.workflow.${workflowID}`, workflow))
+          dispatch(setState(`classifier.tasks.${workflowID}`, workflow.tasks))
+          return resolve()
+        })
+      } else {
+        return resolve()
+      }
+   })
+  }
+}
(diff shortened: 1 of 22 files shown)


src/actions/classifier.js
```diff
@@ -0,0 +1,35 @@
...
+
+export function fetchWorkflow(workflowID) {
+  return (dispatch, getState) => {
+    return new Promise ((resolve) => {
+      if (isNil(getState().classifier.workflow[workflowID])) {
+        return apiClient.type('workflows').get({id: workflowID}).then(([workflow]) => {
+          dispatch(setState(`classifier.workflow.${workflowID}`, workflow))
+          dispatch(setState(`classifier.tasks.${workflowID}`, workflow.tasks))
+          return resolve()
+        })
```
wgranger: I believe there should be a catch here as well. I put in a bogus workflow id into `startNewClassification` to test how it would handle a rejection, and I was getting an unhandled promise rejection with the app hanging on the load screen, probably because there wasn't a catch when this function was called on line 11.

robinschaaf: Ahh OK, good catch!

src/components/ClassificationPanel.js
```diff
@@ -0,0 +1,76 @@
...
+  render() {
+    const tabs =
+      <View style={styles.tabContainer}>
+        <TouchableOpacity
+          style={[styles.tab]}>
+          <StyledText text={ 'QUESTION' } />
+        </TouchableOpacity>
+        <TouchableOpacity
+          style={[styles.tab, styles.deselectedTab]}>
+          <StyledText text={ 'TUTORIAL' } />
```
wgranger: Will this tab/text still show up if a project does not have a tutorial? Although most production projects have a tutorial, it needs to be activated in the project builder first.

wgranger: Oh wait, I see the `hasTutorial` prop, so I'm guessing that handles this question.

robinschaaf: No it won't show up if there isn't one (that part will be included with the upcoming Tutorial PR).

wgranger: Sounds good.

src/components/Question.js
```diff
@@ -0,0 +1,63 @@
...
+export class Question extends Component {
+  render() {
+    return (
+      <View style={[styles.questionContainer, { height: this.props.questionContainerHeight }]}>
+        <View style={styles.question}>
+          <StyledMarkdown
+            extraCSS={ 'p {font-size: 14px; font-weight: 500; margin: 5px 0 0;}' }
+            markdown={ this.props.question }
+            width={ Dimensions.get('window').width - 100 }
+            onResize={ (newHeight) => this.props.setQuestionContainerHeight(newHeight) }
```
wgranger: I'm curious why this `onResize` is needed. Is it for switching in between portrait and landscape? Maybe resizing when the emulator changes sizes? Curious how a resize might be needed when a device has a set window size.

src/components/StyledMarkdown.js
```diff
@@ -0,0 +1,100 @@
...
+
+const WEBVIEW_REF = 'WEBVIEW_REF'
+
+class StyledMarkdown extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { height: 0 }
+  }
+
+  render() {
```
wgranger: Can you place this `render()` after the other functions in this component to keep things consistent?

src/components/Question.js
robinschaaf: Yeah, this is strange.  It turns out that when you use webview you have to hardcode the height of the webview container.  Unfortunately for most everything with markdown we won't know the height until rendered (for example the Question could be one, or a few lines long).  As the result of this, there's a workaround with an event listener on the markdown html template that sets the size, and this is the callback from that.  I'll rename the function though, since by the name `onResize` it's not very obvious that this is what it's doing.

src/components/StyledMarkdown.js
robinschaaf: Yep

src/components/StyledMarkdown.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const resultHTML = defaultHTML
+      .replace('$body', result)
+      .replace('$extraCSS', this.props.extraCSS)
+
+    const displayWidth = this.props.width || Dimensions.get('window').width - 40
+
+    const webviewComponent =
+      <WebView
+        ref={WEBVIEW_REF}
```
wgranger: React changed the way they handle `refs`. Instead of using string-based refs, callbacks are now preferred. There are some good docs from [React Native](https://facebook.github.io/react-native/docs/direct-manipulation.html) and [React](https://facebook.github.io/react/docs/refs-and-the-dom.html).

src/components/StyledMarkdown.js
```diff
@@ -0,0 +1,100 @@
...
+      this.props.onReceivedHeight(htmlHeight)
+    }
+  }
+
+  render() {
+    //Markdownz uses markdown-it-imsize for image sizes, however this package
+    //requires fs, which is a node core module and not available natively
+    //Also, we can't use the image sizes.  This regexp removes them, prior to
+    //being sent to renderer, or else they're interpreted as links
+    let markdown = this.props.markdown.replace(/ =[0-9]+x[0-9]*\)/g, '\)')
```
wgranger: This is kinda crazy, but I suppose it is a viable solution. Did you find this workaround online somewhere? Just curious, since I know we talked about using Markdown in the app briefly not too long ago.

wgranger:
Can you mention briefly what is being done with `default-md-html.js` and `remove-md-tab.js` so we can have some reference of how it is being included in this PR? It's not quite intuitive on first glance what the purpose is and it would be good to have some documentation here.

src/components/StyledMarkdown.js
robinschaaf: Well, I looked at the code being used by markdownz that does the image sizes (https://github.com/edpaget/markdown-it-imsize/blob/master/lib/index.js) and constructed a regexp based on the criteria they use.  (I couldn't use markdownz because it didn't work with react-native babel so I basically tried to reproduce it as much as possible, with markdown-it).  
 I couldn't do this as a markdown plugin because it gets parsed first and was already being turned into a link by the time it made it to the plugin (but since these are images, we want to display the image and not the link).  I can add it to the comments of this file if you think that would be best.

robinschaaf:
Sure thing!  Where do you think would be most appropriate?  In the comments of the code or just in the PR?

wgranger:
I think in the PR would make more sense. It would be more visible as opposed to making a comment in the code.

robinschaaf:
OK, I updated the PR description

src/components/StyledMarkdown.js
wgranger: Wow. Good workaround. I think the comment here is sufficient. I saw there was already a comment in the code.

wgranger:
Alright. Things seem to be in a good place and error handling in working well in both iOS and Android.

PR #113, opened 2017-04-18
Fix user auth error

robinschaaf:
Fixes an issue where users that can't auth will receive an error.  (this appears to just happen in dev when switching environments).  Rather than displaying an error this will redirect to the sign in page if the user isn't logged in.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/actions/user.js
+++ b/src/actions/user.js
@@ -43,13 +43,18 @@ export function loadUserData() {
           dispatch(loadProjectWorkflows()),
         ])
       } else {
-        return Promise.all([
-          dispatch(loadUserAvatar()),
-          dispatch(loadUserProjects()),
-          dispatch(loadNotificationSettings()),
-          dispatch(loadSettings()),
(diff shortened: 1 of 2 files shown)


src/actions/user.js
```diff
@@ -43,13 +43,18 @@ export function loadUserData() {
...
+        dispatch(getAuthUser()).then(() => {
+          return Promise.all([
+            dispatch(loadUserAvatar()),
+            dispatch(loadUserProjects()),
+            dispatch(loadNotificationSettings()),
+            dispatch(loadSettings()),
+            dispatch(loadProjectWorkflows()),
+          ])
+        }).catch(() => {
+          dispatch(setState('errorMessage', ''))
```
wgranger: This just sets an empty error message if one is found?

wgranger: Ah, ok. Seemed odd initially to do this, but if it is to clear out the error so UI is better then this makes sense.

PR #114, opened 2017-04-19
Add Tutorial to Classification Panel

robinschaaf:
Adds the Tutorial and supporting items:
  *  Sets up the project preferences (for recording tutorial completion) in case the user hasn't visited the project before on the desktop
  *  Alters the ActivePreferences to store all projects returned in redux, and then changes the Recent Projects list and CircleRibbon to filter only active
  *  Gets the tutorial and any attached images (media)
  *  Marks tutorial as completed.  For guests it will remember that they completed it for that session.
  *  Adds a SizedImage component that does animations on images when loaded in the Tutorial.  This helps keep it from being jumpy.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/actions/classifier.js
+++ b/src/actions/classifier.js
@@ -1,14 +1,30 @@
 import apiClient from 'panoptes-client/lib/api-client'
-import { isNil } from 'ramda'
+import { forEach, isEmpty, isNil } from 'ramda'
 import { setState } from '../actions/index'
+import { getAuthUser } from '../actions/auth'
 import { Actions } from 'react-native-router-flux'
 import { Alert } from 'react-native'
 
 export function startNewClassification(workflowID) {
-  return (dispatch) => {
+  return (dispatch, getState) => {
+    dispatch(setState('classifier.isFetching', true))
     dispatch(setState('loadingText', 'Loading Workflow...'))
     dispatch(setState('classifier.currentWorkflowID', workflowID))
     dispatch(fetchWorkflow(workflowID)).then(() => {
+      if (getState().classifier.tutorial[workflowID] !== undefined) {
+        return
+      }
+      return dispatch(fetchTutorials(workflowID))
+    }).then(() => {
+      if (getState().classifier.project[workflowID] !== undefined) {
+        return
+      }
+      return dispatch(fetchProject(workflowID))
+    }).then(() => {
+      return dispatch(setupProjectPreferences(workflowID))
+    }).then(() => {
+      return dispatch(setNeedsTutorial())
+    }).then(() => {
       dispatch(setState('classifier.isFetching', false))
     }).catch(() => {
       Alert.alert('Error', 'Sorry, but there was an error loading this workflow.  Please try again later.',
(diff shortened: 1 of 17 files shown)


src/actions/classifier.js
```diff
@@ -1,14 +1,30 @@
...
+      }
+      return dispatch(fetchProject(workflowID))
+    }).then(() => {
+      return dispatch(setupProjectPreferences(workflowID))
+    }).then(() => {
+      return dispatch(setNeedsTutorial())
+    }).then(() => {
       dispatch(setState('classifier.isFetching', false))
     }).catch(() => {
       Alert.alert('Error', 'Sorry, but there was an error loading this workflow.  Please try again later.',
```
wgranger: When I am signed in, I keep getting to this `.catch()` when trying to open Planet Four. However, there is no problem when I am signed out. It could be an issue with my environment though?

src/actions/classifier.js
```diff
@@ -1,14 +1,30 @@
...
+    }).then(() => {
+      if (getState().classifier.project[workflowID] !== undefined) {
+        return
+      }
+      return dispatch(fetchProject(workflowID))
+    }).then(() => {
+      return dispatch(setupProjectPreferences(workflowID))
+    }).then(() => {
+      return dispatch(setNeedsTutorial())
+    }).then(() => {
```
wgranger: When I am signed in, I keep getting to this .catch() when trying to open Planet Four. However, there is no problem when I am signed out. It could be an issue with my environment though?

robinschaaf: Hmm, might not be environment.  Have you gone to P4 ridges on desktop before?  If you haven't then it may not be creating the project preferences correctly.  I had this fixed in the beta version of the app but might have missed something in the transition.

wgranger: I have visited Planet Four on desktop before. I also have the `tutorial_completed_at` tag in my preferences, for what it's worth.

robinschaaf: Ahh, ok good info.  Then that's not the problem.  I'll look into it.

robinschaaf: OK, looks like I messed something up in a rebase or something.  Should be working again.

src/actions/classifier.js
```diff
@@ -35,3 +51,143 @@ export function fetchWorkflow(workflowID) {
...
+    const workflow = getState().classifier.workflow[workflowID]
+    const projectID = workflow.links.project
+
+    return new Promise ((resolve, reject) => {
+      if (getState().user.isGuestUser){
+        return resolve()
+      }
+
+      dispatch(getAuthUser()).then((userResource) => {
+        return userResource
```
wgranger: Am I missing something here? It seems a bit odd to have a repeating `.then(userResource)`.

robinschaaf: Whoops.  Fixed it.

src/actions/classifier.js
```diff
@@ -35,3 +51,143 @@ export function fetchWorkflow(workflowID) {
...
+          }
+
+          const projectPreference = {
+            links: { project: projectID },
+            preferences: {}
+          }
+          apiClient.type('project_preferences').create(projectPreference).save().then((projectPreferenceResource) => {
+            return projectPreferenceResource
+          }).then((projectPreferenceResource) => {
+            projectPreferenceResource.get('project').then((project) => {
```
wgranger: Is the project state not already fetched and set in line 22 of classifier.js?

src/actions/classifier.js
```diff
@@ -35,3 +51,141 @@ export function fetchWorkflow(workflowID) {
...
+          if (projectPreferences) {
+            return resolve()
+          }
+
+          const projectPreference = {
+            links: { project: projectID },
+            preferences: {}
+          }
+          apiClient.type('project_preferences').create(projectPreference).save().then((projectPreferenceResource) => {
+            return projectPreferenceResource
```
wgranger: Same thing here as before. I believe you have a repeating `.then((projectPreferenceResource)`.

src/actions/classifier.js
wgranger: Oh, it appears state is being set here a bit differently than above.

robinschaaf: Oh - yes good call!

src/actions/classifier.js
robinschaaf: Was able to remove altogether after above comment.

src/actions/classifier.js
robinschaaf: I think you were right - they're both for the current workflow's project so it works to remove it and I just get the display name and slug from the stored project instead.

src/actions/classifier.js
```diff
@@ -35,3 +51,139 @@ export function fetchWorkflow(workflowID) {
...
+      userResourse.get('project_preferences', {project_id: projectID}).then (([projectPreferences]) => {
+        if (!projectPreferences.preferences.tutorials_completed_at) {
+          projectPreferences.preferences.tutorials_completed_at = {}
+        }
+        const completed = {
+          [tutorialID]: now
+        }
+        projectPreferences.update({
+          preferences: {
+            tutorials_completed_at: completed
```
wgranger: I'm a bit concerned creating a new object `completed` (above) and setting `tutorials_completed_at` to that object might overwrite any other tutorial already present in `preferences.tutorial_completed_at` (ex. if there is more than one tutorial). I'd just double check this, because my logic could be wrong in examining this.

robinschaaf: Sure.  Let me double check that - I'm not sure that I tested with more than one tutorial.

srallen: They're tracked by tutorial id on PFE since tutorials belong to workflows: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/tutorial.cjsx#L114-L115

robinschaaf: Yes, the tutorials_completed_at object was being overwritten (it was using the tutorial ID key, but the entire object was overwritten).  Should be fixed now.

src/components/SwipeClassifier.js
```diff
@@ -2,50 +2,94 @@ import React from 'react'
...
+          setQuestionVisibility = { this.setQuestionVisibility }>
+          { classifierOrTutorial }
         </ClassificationPanel>
+
+      return (
+        this.props.needsTutorial ? tutorial : classificationPanel
       )
     }
 
+    const renderSpinner = this.props.isFetching || this.props.workflow === undefined
```
wgranger: Does this mean if there is no `props.workflow` the spinner will render indefinitely, or is a call made somewhere else to grab the workflow? Seems like bad UI if it's the former.

robinschaaf: It ought not get there because of that catch which give an alert error and sends them back, but I don't think that was needed at any rate  (I added it while testing something else)  so I've removed it.

PR #115, opened 2017-04-19
Add Task Help

robinschaaf:
Add task help to display a question mark icon next to the question.  It opens the task help in a modal.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/Question.js
+++ b/src/components/Question.js
@@ -5,6 +5,7 @@ import {
 } from 'react-native';
 import EStyleSheet from 'react-native-extended-stylesheet'
 import StyledMarkdown from './StyledMarkdown'
+import TaskHelp from './TaskHelp'
 import { connect } from 'react-redux'
 import { setState } from '../actions/index'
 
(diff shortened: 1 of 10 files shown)


wgranger:
This PR looks pretty solid, and 75% of it seems to be snapshots taken for testing. However, I think the close button should either change colors or change position. Consider the two photos. It would be nice if the modal never raised above the close button, so the button is always visible over the navbar. Currently, when you scroll down the X can become hidden and seems to disappear with a white background. 
screen shot 2017-04-20 at 2 27 37 pm
screen shot 2017-04-20 at 2 26 13 pm

robinschaaf:
Oh, yes that makes sense.  I'll take a look at it.

wgranger:
Also, I'd double check how it appears on Android. I've tried a couple emulators, and the close button is cut off at the bottom on both. 
screen shot 2017-04-20 at 2 42 58 pm

robinschaaf:
OK, this should be fixed now, I wasn't having problems in Android but the fix may have solved that.  Let me know if you still see it.

wgranger:
Nice. That commit fixed both issues.

PR #116, opened 2017-04-21
Add swipe classifier and classification [WIP] [RFC]

robinschaaf:
Adds three new components:
  *  SwipeClassifier: entry point for swipe classification
  *  SwipeSubject:  displays subject image, animates entrance for front/back subjects, adds Already Seen
  *  Swipeable: The swipe handling functionality (the overlay and animations)

Adds new actions / data to store:
  *  classifier: create and save annotations and classifications, track subjects seen
  *  device: store device size information and the subject displays sizes
  *  subject: subject queue fetch, subject load for current/next subject, set subject image sizes

Coming in subsequent PRs:
  *  Google Analytics tracking
  *  Image Zoom
  *  Bottom tabs (Yes/No buttons)
  *  Field Guide

# Review Checklist

- [x] Are tests passing?


src/actions/classifier.js
```diff
@@ -25,29 +25,106 @@ export function startNewClassification(workflowID) {
...
       return dispatch(setNeedsTutorial())
     }).then(() => {
+      dispatch(setState('loadingText', 'Loading Subjects...'))
+      return dispatch(loadSubjects())
+    }).then(() => {
+      return dispatch(setSubjectsToDisplay())
+    }).then(() => {
+      //now we can create the first classification!!
+      let subject = getState().classifier.subject[workflowID]
+      let workflow = getState().classifier.workflow[workflowID]
```
wgranger: This is an above `let` can be `const` since the variables aren't redeclared.

src/actions/classifier.js
```diff
@@ -25,29 +25,106 @@ export function startNewClassification(workflowID) {
...
+          workflow: workflow.id,
+          subjects: [subject.id]
+        }
+      })
+    }).then((classification) => {
+      dispatch(setState(`classifier.classification.${workflowID}`, classification))
       dispatch(setState('classifier.isFetching', false))
-    }).catch(() => {
+    }).catch((e) => {
+      console.log('Error in catch! ', e)
```
wgranger: Is this necessary? I believe this would be the only `console.log` in the app, and I'm not sure where it would be visible.

src/actions/classifier.js
robinschaaf: Sure

src/actions/classifier.js
robinschaaf: Yeah I don't think it is necessary, it is helpful with debugging (I was thinking for the future) but that's the only purpose

src/actions/subject.js
```diff
@@ -0,0 +1,110 @@
+import apiClient from 'panoptes-client/lib/api-client'
+import { isNil, length, prepend } from 'ramda'
+import { setState } from '../actions/index'
+import { Image } from 'react-native'
+import getSubjectLocation from '../utils/get-subject-location'
+
+export function fetchUpcomingSubjects(workflowID) {
+  return dispatch => {
+    return new Promise ((resolve, reject) => {
+      apiClient.type('subjects').get({workflow_id: workflowID, sort: 'queued'}).then((subjects) => {
```
wgranger: I'm curious if this `sort: 'queued'` is working. The app queue seems to work fine when I test it in iOS and Android, so it might be fine. However, I don't see in the Panoptes serializer that you can sort subjects by 'queued', and there is also a subject_queue route. Another set of eyes on this would be good.

zwolf: `sort: "queued"` results in the same request as using `/api/subjects/queued`.  

See https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/subjects_controller.rb#L13

src/components/Swipeable.js
```diff
@@ -0,0 +1,221 @@
...
+              />
+              <Animated.View style={[styles.overlayContainer, leftOverlayStyle, imageSizeStyle]} />
+              <Animated.View style={[styles.overlayContainer, leftOverlayTextStyle, imageSizeStyle]}>
+                <StyledText additionalStyles={[styles.answerOverlayText]} text={ answers[0].label } />
+              </Animated.View>
+
+              <Animated.View style={[styles.overlayContainer, rightOverlayStyle, imageSizeStyle]} />
+              <Animated.View style={[styles.overlayContainer, rightOverlayTextStyle, imageSizeStyle]}>
+                <StyledText additionalStyles={[styles.answerOverlayText]} text={ answers[1].label } />
+              </Animated.View>
```
wgranger: Am I correct in assuming both the blue/red containers and also the yes/no text is _always_ present over the subject, but their visibility is determined by the swipe movement left or right?

src/components/Swipeable.js
```diff
@@ -0,0 +1,221 @@
...
+
+Swipeable.propTypes = {
+  subject: React.PropTypes.shape({
+    id: React.PropTypes.string
+  }),
+  nextSubject: React.PropTypes.shape({
+    id: React.PropTypes.string
+  }),
+  subjectSizes: React.PropTypes.object,
+  seenThisSession: React.PropTypes.array,
```
wgranger: Let's be a bit more specific here, as we've been doing in PFE. Use `React.PropTypes.arrayOf(React.PropTypes.string/object/etc.)`. This is preferred as `React.PropTypes.array` can be a bit ambiguous. [Here](https://facebook.github.io/react/docs/typechecking-with-proptypes.html) is some info from the React docs.

src/components/Swipeable.js
robinschaaf: Yes that's correct - the opacity for each changes depending on where the user is panning, e.g. `let opacityRight = pan.x.interpolate({inputRange: [0, 100, 200], outputRange: [0, .6, .8]})`

src/components/Swipeable.js
robinschaaf: Sure thing.

src/components/Swipeable.js
```diff
@@ -0,0 +1,221 @@
...
+        Animated.spring(this.state.pan, {
+          toValue: {x: 0, y: 0},
+          friction: 4
+        }).start()
+      }
+    }
+    })
+  }
+
+  onAnswered = (answer) => {
```
wgranger: Is there a reason this is using the arrow function? It's not really a big deal and I'm probably being too pedantic but was wondering if it could be written like the other functions.

PR #117, opened 2017-04-25
Add Field Guide

robinschaaf:
Adds Field Guide to swipe interface
  *  Fetches the project's Field Guide and stores it in redux
  *  Adds a new component for SwipeTabs - which will display the Yes/No buttons at the bottom, with the Field Guide in between.  The Y/N buttons will always appear, so if there's no Field Guide, the gray box will still render, but there won't be a Field Guide button (Y/N buttons will be in a later PR)
  *  Field Guide will change height depending on the height of the content (from the StyledMarkdown content).  The user can drag it to be higher - if the component is tall enough - but only to a max of 60% of the screen.  The user can also drag it to be lower.
  *  This also modifies the StyledText component to accept props for an ellipsis, used on the Field Guide item row in case the title is too long

# Review Checklist

- [x] Are tests passing?


src/actions/classifier.js
```diff
@@ -25,6 +25,11 @@ export function startNewClassification(workflowID) {
     }).then(() => {
       return dispatch(setNeedsTutorial())
     }).then(() => {
+      if (getState().classifier.guide[workflowID] !== undefined) {
+        return
+      }
+      return dispatch(fetchFieldGuide(workflowID))
+    }).then(() => {
```
wgranger: Is it possible to combine all of these `.then()` statements rather than having a separate call and return for each? Might save a few lines of code and clean up this component a bit.

robinschaaf: I was thinking of putting the returns actually in the call, is that what you're talking about?  Also some of them are dependent on others,  like tutorial is dependent on project preferences.  Or not sure what you're referring to otherwise.

wgranger: Oh. I was thinking, for example, that lines 22, 24, and 26 could be in the same `.then()` block, but if they are async/dependent on each other and need to be chained, that's fine.

robinschaaf: Oh, got it.  The ones that aren't dependent could get placed together, I'll try that.

robinschaaf: Well, it turns out there isn't much that doesn't have a dependent, just the Field Guide and setNeedsTutorial (which actually needs to be loaded before the screen loads so that it goes straight to the tutorial rather than opening the classifier and then moving to the tutorial) so it didn't seem to buy much.

src/components/FieldGuide.js
```diff
@@ -0,0 +1,297 @@
...
+        easing: Easing.linear,
+        duration: duration,
+      }
+    ).start()
+  }
+
+  setHeight(height) {
+    const newHeight = height + this.state.headerHeight
+    this.setState({height: newHeight})
+    this.animateHeight(newHeight < MAX_HEIGHT ? newHeight : MAX_HEIGHT)
```
wgranger: I believe this is the part of the code that keeps the field guide at 33 when the guide shows icon rows. My first inclination when opening the field guide was to raise it to see all items, but the guide kept snapping to the `MAX_HEIGHT`. Is there a reason we are keeping a `MAX_HEIGHT` when in icon view? I imagine it is to annotate while looking at the field guide, but there may be an instance where the user wants to expand the guide to see more rows. Worth a convo regardless.

src/components/FieldGuide.js
```diff
@@ -0,0 +1,297 @@
...
+    position: 'absolute',
+    bottom: -25,
+    right: 15,
+    backgroundColor: 'transparent',
+  },
+})
+
+FieldGuide.propTypes = {
+  guide: React.PropTypes.shape({
+    items: React.PropTypes.arrayOf(React.PropTypes.shape({
```
wgranger: Saying `arrayOf(React.PropTypes.object)` is probably specific enough here, but this is fine for now.

src/components/FieldGuideItemDetail.js
```diff
@@ -0,0 +1,76 @@
...
+  ScrollView,
+  View
+} from 'react-native'
+import EStyleSheet from 'react-native-extended-stylesheet'
+import StyledMarkdown from './StyledMarkdown'
+import StyledText from './StyledText'
+import SizedImage from './SizedImage'
+import Button from './Button'
+
+const ITEM_ICON_HEIGHT = 100
```
wgranger: Changing this to `ITEM_ICON_RADIUS` might make more sense for when it is used below in the styles.

src/components/FieldGuide.js
robinschaaf: It should display all items (and automatically become higher if there are more items), but to the max of 60% of the screen (which is admittedly arbitrary, happy to change it if you think it makes sense).  It should be displaying all items,however.  Is that not what you're seeing?  (I just added some more in the field guide)

image

src/components/FieldGuideItemDetail.js
robinschaaf: Sure

src/components/FieldGuide.js
wgranger: Got it. Just noticed you added those items, which changed the functionality a bit, so this makes sense now.

src/constants/mobile_projects.js
```diff
@@ -1,6 +1,10 @@
-export const MOBILE_PROJECTS = ['21', '55', '60', '2299', '347', '1338', '36', '1457', '17', '2455', '2492', '2977', '3219', '2501']
+export const MOBILE_PROJECTS = ['60', '2299', '347', '1338', '36', '1457', '17', '2455', '2492', '3219', '3321', '2501', '14', '1558']
 
 export const SWIPE_WORKFLOWS = [
+  { projectID: '3321', workflowID: '2725' },
+  { projectID: '3321', workflowID: '3437' },
```
wgranger: Did you mean to introduce this test project into the PR? (Scary Bugs)

wgranger:
I'm noticing an error after a recent commit/rebase. Whenever I open a project (Scary Bugs) and go into another mobile friendly project (Planet Four) I get this error. 
screen shot 2017-04-25 at 2 42 29 pm

src/constants/mobile_projects.js
robinschaaf: Nope, I'll remove it

robinschaaf:
Yeah that's a bug I opened an issue for earlier today #118  , I have a branch up to fix it and will be opening a PR soon.  In the meantime you'll need to refresh each time you want to navigate to a new project, unfortunately.

wgranger:
Ugh. Yes, best to get that fixed ASAP.

wgranger:
I've finished my initial pass. The last item I'll mention is that the field guide should have a max height that keeps it from scrolling up so high it is hidden behind the navbar. I did that once on accident while testing, and the only way I could get rid of the navbar was by going back. Perhaps the field guide should only be expanded to the extent of the image below.
screen shot 2017-04-25 at 3 41 48 pm

wgranger:
This is looking good.

PR #119, opened 2017-04-25
Move first task into redux to prevent errors

### LINKED ISSUE #118: IsFetching not working when navigating between multiple swipe workflows
IsFetching should be dependent on the workflow, otherwise it initializes with the wrong value.

robinschaaf:
Fixes #118 

Add the first task to redux store to be able to utilize initialize state - also helps with keeping the logic to determine the first task in the action.

# Review Checklist

- [x] Are tests passing?


wgranger:
This will need a rebase after merging the field-guide and classifier.

PR #122, opened 2017-04-27
Switch projects to fetch from db

robinschaaf:
Switch mobile friendly projects to fetch from DB rather than being hardcoded.
  *  Uses the project.mobile_friendly attribute to fetch the projectList
  *  Uses the intersection of user.activeProjects and project.mobile_friendly to populate the Recents list
  *  Fetches the workflows for the mobile_friendly projects 
  *  Determines if the workflow is swipe friendly by checking that configuration.swipe_enabled is true 
and that the workflow adheres to the valid swipe workflow criteria

# Review Checklist

- [x] Are tests passing?


wgranger:
Can you rebase this so that bug fix PR is included?

robinschaaf:
Sure

src/reducers/__tests__/__snapshots__/reducers-test.js.snap
```diff
@@ -26,10 +26,12 @@ Object {
   "notifications": Object {
     "general": true,
   },
-  "projectList": Object {},
-  "projectWorkflows": Object {},
+  "projectList": Array [],
+  "projectListHolding": Array [],
```
wgranger: What is the purpose of `projectListHolding`? You mentioned it briefly above, but I was curious in general.

robinschaaf: It's just for UI.  While the projectList is being fetched it has a lot of promises and if I were directly modifying projectList it would remove stuff and then put it back, resulting in a jumpiness.  This way I just wait until everything's fetched and then reassign the projectList at once then.

wgranger: Ahh, ok. That explains the line comment in the other component.

wgranger:
This is good to go. Mobile_friendly projects will need to be updated individually so they appear in the app, but this code works.

PR #123, opened 2017-04-28
Set to allow portrait only

robinschaaf:
Add restriction to iOS and Android to allow portrait mode only, since the swipe projects will need to be displayed in portrait and we have to set allowable mode at the app level.

# Review Checklist

- [x] Are tests passing?


wgranger:
This appears to work for Android, but my screen is still rotating in the iOS emulator.

robinschaaf:
Darn it, I swear that worked.  I ended up having to remove react-native-orientation altogether, which ended up not being needed at all after this change.

wgranger:
Ok, this seems to be working on iOS/Android now.

PR #124, opened 2017-04-28
Add Y/N buttons

robinschaaf:
Add Yes and No buttons with swipe icons to the bottom of the screen to left and right of the Field Guide.  Also moves onAnswered function and answers up to parent SwipeClassifier so that the SwipeTabs has access to it.

- [x] Are tests passing?


wgranger:
This will need a rebase after merging the image expander.

src/components/SwipeClassifier.js
```diff
@@ -73,6 +89,8 @@ export class SwipeClassifier extends React.Component {
     const renderClassifierOrTutorial = () => {
       const key = this.props.workflow.first_task //always just one task
       const task = this.props.workflow.tasks[key]
+      const answers = reverse(task.answers)  //Yes is listed first in project, but we need No listed first (on left)
```
wgranger: Does this reflect what we are asking researchers to do on the project builder? (making "yes" the first answer and "no" the second answer)

robinschaaf: Yeah it says "If you have a Yes/No question, we recommend Yes is the first option listed so that it appears on the right."  This is because the existing projects (Steller Sea Watch and P4:ridges) had listed Yes first, I think because it looks better in PFE like that.  So then the swipe app has to reverse that order.  (I think you can access the Insect Classification csv export, and you can check the resulting answers in that)

wgranger: Ok, good. Just wanted to make sure this was clear somewhere to the project creator.

PR #125, opened 2017-04-28
Add Unlinked Tasks

robinschaaf:
Add Unlinked Tasks/Shortcuts and ability to save additional annotations for the unlinked task and the ability to remove annotations in case a the switch is "unchecked".  This will allow more than two unlinked tasks, but other checks should prevent a workflow with more than two from being swipe eligible, due to screen size.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/actions/classifier.js
+++ b/src/actions/classifier.js
@@ -1,6 +1,6 @@
 import apiClient from 'panoptes-client/lib/api-client'
 import { forEach, isEmpty, isNil, map, remove, toPairs } from 'ramda'
-import { addState, setState } from '../actions/index'
+import { addState, removeState, setState } from '../actions/index'
 import { Actions } from 'react-native-router-flux'
 import { Alert, Platform } from 'react-native'
 import { getAuthUser } from '../actions/auth'
@@ -69,6 +69,13 @@ export function saveAnnotation(task, value) {
   return (dispatch, getState) => {
     const workflowID = getState().classifier.currentWorkflowID
     dispatch(setState(`classifier.annotations.${workflowID}.${task}`, value))
+  }
+}
+
+export function removeAnnotationValue(task, value) {
+  return (dispatch, getState) => {
+    const workflowID = getState().classifier.currentWorkflowID
+    dispatch(removeState(`classifier.annotations.${workflowID}.${task}`, value))
   }
 }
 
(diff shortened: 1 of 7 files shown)


wgranger:
This will need a rebase after merging other branches.

src/actions/classifier.js
```diff
@@ -72,6 +72,13 @@ export function saveAnnotation(task, value) {
   }
 }
 
+export function removeAnnotationValue(task, value) {
+  return (dispatch, getState) => {
+    const workflowID = getState().classifier.currentWorkflowID
+    dispatch(removeState(`classifier.annotations.${workflowID}.${task}`, value))
```
wgranger: Does this remove the original annotation value (yes or no) and replace it with the shortcut value? The shortcut component was recently updated to retain the original annotation/answer and add that unlinked task as an additional answer.

src/components/SwipeClassifier.js
```diff
@@ -83,6 +89,15 @@ export class SwipeClassifier extends React.Component {
     this.props.saveThenStartNewClassification()
   }
 
+  onUnlinkedTaskAnswered = (task, value) => {
+    const taskAnnotations = this.props.annotations[task] || []
+    if (contains(value, taskAnnotations)) {
+      this.props.removeAnnotationValue(task, value)
```
wgranger: Yes, this _might_ change depending on the comment above.

src/actions/classifier.js
robinschaaf: Oh, do you mean there can only be one value allowed at a time?  It allows multiples right now.  For example like this: (Nothing Here, nothing Here either are two unlinked tasks)
[{"task":"T1","task_label":"Is this a **scary** bug?","value":["Nothing Here","Nothing Here either"]},{"task":"T0","task_label":"Is this a **scary** bug?","value":"No"}]

wgranger: Not quite. Let's say the user selected "Nothing here" and then answers "yes" to the question. Will the final annotation have both "yes" and "Nothing here" saved as an answer.

robinschaaf: Yeah, "Nothing here" under T1, and "yes" under T0.

wgranger: Perfect, that is what should happen. Initially, the shortcut task was set to cancel out the answer to the original question, but it seems it is set up correctly here.

PR #127, opened 2017-05-16
Don't Initiate Tutorial if Non-existent 

wgranger:
Describe your changes.
Snapshots at Sea was not opening in the native app's swipe classifier. This turned out to be an issue because the `SwipeClassifier` component was trying to initialize a `Tutorial` component when the project did not have a tutorial.

# Review Checklist

- [ ] Are tests passing?

### DIFF
--- a/src/components/SwipeClassifier.js
+++ b/src/components/SwipeClassifier.js
@@ -185,7 +185,7 @@ export class SwipeClassifier extends React.Component {
 
       //needsTutorial is for the first time a guest or user visits this project
       return (
-        this.props.needsTutorial ? tutorial : classificationPanel
+        this.props.needsTutorial && !isEmpty(this.props.tutorial) ? tutorial : classificationPanel
       )
     }
 


wgranger:
You know, that variable is a bit misleading. It appears that boolean reflects if it's the first time classifying on a project and if the tutorial should pop up on the first viewing. I could change that variable a bit for clarity. I'm not sure off the top of my head how many components `needsTutorial` touches. Perhaps simply changing the name to something like `firstTimeView` or something of the sort would be more appropriate.

eatyourgreens:
Good point. If it's used elsewhere to test for a first visit, then the name should reflect that.

wgranger:
Looking at the code again, I decided to keep the variable name, but set the variable to false if a tutorial doesn't exist. It looks like this was originally the intention of the code, but `!getState().classifier.tutorial[workflowID]` was always being set to `false` since `classifier.tutorial[workflowID]` would default to an empty object when a tutorial isn't present.

eatyourgreens:
Looks good to me.

Not a blocking issue, but will tests catch the bug if it regresses?

wgranger:
@eatyourgreens If I understand your above comment correctly, there should have initially been a test that caught this bug in the first place. We should add a test that will be triggered to catch the bug that currently exists in the code. 

If that is the case, I can add a test to imitate current functionality and make sure the bug stays dead.

eatyourgreens:
@wgranger that's what I was thinking. The test could come in another PR, unless it's a quick one to write.

wgranger:
I'll take a stab at it today after standup. If I can get one written quickly then I'll push it onto this branch. If it's looking to take longer, I'll let you know and it can come in another PR. Hopefully it should be quick and easy though.

wgranger:
This is looking like it will take a while to figure out because jest has an issue with easily being able to change a mock (ultimately the state) between tests. There is some more info [here](https://github.com/facebook/jest/issues/2577). It's going to take a bit more digging to figure out the best way to do this cleanly without creating a separate file that sets up the classifier the exact same way (sans tutorial for the purpose of this test). Creating a similar file seems redundant and unnecessary, so digging a bit more is preferable to find the best solution to keep the test in the same `SwipeClassifier-test.js` file.

eatyourgreens:
No problem. Happy to merge this as-is.

PR #132, opened 2017-08-14
Change Swipe Tag Location

wgranger:
Describe your changes.
This PR changes the location of the swipe_classifier tag with the discussion made in [this](https://github.com/zooniverse/Panoptes/issues/2432) backend issue.

# Review Checklist

- [x] Are tests passing?


wgranger:
We are no longer changing the location of the `swipe_enabled` tag on the backend. Putting `swipe_enabled` on the workflow resource would be a bit too specific for our purposes. Instead, expect a PR up momentarily to adjust the project query to retrieve only the `launch_approved` projects we want.

PR #134, opened 2017-12-07
Restyle UnlinkedTask Component

wgranger:
There's an issue with the way unlinked tasks/shortcuts are displayed in the mobile app (see images attached). This clears up the issue and should be merged in order to be live by the Dec. 13 launch of Floating Forests (with a couple days needed for Apple to approve the update).

Describe your changes.
Changed styles for the UnlinkedTask component.

# Review Checklist

- [x] Are tests passing?

image uploaded from ios
image1

### DIFF
--- a/src/components/UnlinkedTask.js
+++ b/src/components/UnlinkedTask.js
@@ -47,7 +47,7 @@ const UnlinkedTask = (props) => {
 const styles = EStyleSheet.create({
   container: {
     alignSelf: 'flex-start',
-    margin: 10,
+    margin: 15,
     position: 'absolute',
     bottom: 65
   },
--- a/src/components/UnlinkedTask.js
+++ b/src/components/UnlinkedTask.js
@@ -47,23 +47,25 @@ const UnlinkedTask = (props) => {
 const styles = EStyleSheet.create({
   container: {
     alignSelf: 'flex-start',
-    margin: 20
+    margin: 10,
+    position: 'absolute',
+    bottom: 65
   },
   rowContainer: {
     flexDirection: 'row',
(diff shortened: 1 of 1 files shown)


src/components/UnlinkedTask.js
```diff
@@ -47,23 +47,25 @@ const UnlinkedTask = (props) => {
 const styles = EStyleSheet.create({
   container: {
     alignSelf: 'flex-start',
-    margin: 20
+    margin: 15,
+    position: 'absolute',
```
camallen: isn't absolute positioning a bad idea on different view ports?

wgranger: Yes, I think you are right on the absolute positioning. I arrived at this solution because there appears to be some wonkiness with where the unlinked tasks render depending on the image size that I couldn't quite nail down while hacking away yesterday. Unfortunately, a fix is needed by Monday to get an Apple review and release out in time.

I will double check the rendering on several more device screen sizes. iPhone SE/6 and Nexus 5 appeared well according to this PR.

wgranger: Layout appears to be good on iPhones 5-7 and various size Nexus screens.

camallen: If you are happy with it and it's needed then we can revisit it once it's out.

rogerhutchings:
I tried to spin this up locally, but unfortunately didn't have any mobile-friendly projects available, so was unable to test. But I agree with @camallen - if you think it's good to go, we can always go back to it after it's released.

wgranger:
Agreed. I think this is good to go but can always revisit it again after the release.

PR #150, opened 2018-02-12
React native upgrade

### LINKED ISSUE #136: Update react-native version
We are pretty far behind on react native versions. We should try to get as updated as possible

noahmalmed:
Fixes #136  .

Describe your changes.
- Updating react native version
- Updating all packages that depend on react native/ react version

Just a note on this PR: 
I would suggest going this commit by commit. A lot of the code changes are config changes done automatically by the upgrade script. A large portion of the code changes are from `prop-types` being moved into it's own package and out of reacts. I did my best to filter these big, "noisy" changes into their own commits.

# Review Checklist

- [x] Are tests passing?


.gitignore
```diff
@@ -25,25 +25,41 @@ main.jsbundle.meta
...
 local.properties
 zooniverse-key.keystore
 gradle.properties
+*.iml
 
 # node.js
 #
 node_modules/
 npm-debug.log
+yarn-error.log
```
wgranger: Do we need this? I don't believe we are using Yarn.

__mocks__/mockUtils.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import PropTypes from 'prop-types';
+
+// Create a mock component
+export const createMockComponent = function (name) {
+    return class extends React.Component {
+      static displayName =name
```
wgranger: Pedantic, but can you put a space in here?

__mocks__/react-native-drawer.js
```diff
@@ -0,0 +1,4 @@
+import { createMockComponent } from './mockUtils'
+
+export const Drawer = createMockComponent('Drawer');
+export default Drawer;
```
wgranger: It seems odd having two exports here. I think you can remove the first.

__mocks__/react-native-image-pan-zoom.js
```diff
@@ -0,0 +1,4 @@
+import { createMockComponent } from './mockUtils'
+
+export const ImageZoom = createMockComponent('ImageZoom');
+export default ImageZoom;
```
wgranger: Same as above.

wgranger:
Take a look at a project with shortcuts (Floating Forests) on production through iOS and Android. The placement may be off in Android, and it doesn't appear that the text is appearing for either.

index.js
```diff
@@ -10,19 +11,30 @@ import theme from './src/theme'
...
   render() {
+    
     return (
       <View style={styles.container}>
-        <StatusBar
-          barStyle="light-content"
-          hidden={false}
-          translucent={false}
-        />
+        <PlatformSpecificStatusBar />
```
wgranger: Is there a reason this component only renders for iOS?

noahmalmed: Not sure, it was basically the only difference between the two index files before I consolidated them.

wgranger: Ah, gotcha. Didn't notice this used to be the iOS index.

wgranger:
Ok. Things look nice and updated (especially looking at the package.json. However, there is a blocking comment above with the shortcuts. Also, I noted on a few items I'm curious about. It was good to have this spin up without issues first try.

.gitignore
noahmalmed: yeah, we can remove this. I think the update script added it.

noahmalmed:
@wgranger changes made per your request. Everything should look good now. I actually also noticed that we weren't setting the color correctly for the toggle as well as not showing the text. Both of those should be fixed now.

wgranger:
Cool. This LGTM.

PR #152, opened 2018-02-13
Issue151 - Use mobile_friendly workflow attribute

### LINKED ISSUE #151: update workflow request
https://github.com/zooniverse/mobile/blob/master/src/actions/index.js#L148

update the workflow request from:
```
return project.get('workflows', {page_size: 100, active: true})
```
to:
```
return project.get('workflows', {active: true, fields: 'display_name,configuration', mobile_friendly: true})
```

noahmalmed:
Fixes #151.

Describe your changes.
Instead of pulling the top 100 projects from a page, we are pulling only the ones that are `mobile_friendly`.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -145,7 +145,7 @@ export function fetchProjects(parms, stateKey) {
             const promise = apiClient.type('avatars').get(project.links.avatar.id).then((avatar) => {
               project.avatar_src = avatar.src
             }).then(() => {
-              return project.get('workflows', {page_size: 100, active: true})
+              return project.get('workflows', {mobile_friendly: true, active: true})
             }).then((workflows) => {
               project.workflows = tagSwipeFriendly(workflows)
               return dispatch(addState(stateKey, project))
@@ -168,7 +168,7 @@ export function fetchProjects(parms, stateKey) {
 
 function tagSwipeFriendly(workflows) {
   return map((workflow) => {
-    workflow.swipe_verified = !!workflow.configuration.swipe_enabled && isValidSwipeWorkflow(workflow)
+    workflow.swipe_verified = workflow.mobile_friendly && isValidSwipeWorkflow(workflow)
     return workflow
   }, workflows)
 }


mcbouslog:
Noting we should update https://github.com/zooniverse/how-to-zooniverse/blob/master/ProjectLifecycle/mobile.md once ready for merge.

src/actions/index.js
```diff
@@ -168,7 +168,7 @@ export function fetchProjects(parms, stateKey) {
 
 function tagSwipeFriendly(workflows) {
   return map((workflow) => {
-    workflow.swipe_verified = !!workflow.configuration.swipe_enabled && isValidSwipeWorkflow(workflow)
+    workflow.swipe_verified = workflow.mobile_friendly && isValidSwipeWorkflow(workflow)
```
wgranger: I think there still needs to be a check here for `workflow.configuration.swipe_enabled`. It can either be here or in the `isValidSwipeWorkflow()` function. That is, just because a workflow is `mobile_friendly`doesn't necessarily mean it will be `swipe_enabled`. In the future, as the app expands in functionality, we may have a workflow that is `mobile_friendly` but not necessarily `swipe_enabled`, and just because a project meets all the conditions to be a swipe workflow doesn't mean the project owner would want it to be considered such.

Open to debate on this though.

wgranger:
Overall, this looks good and is a needed fix. However, there is just one comment to sort out before merge.

src/actions/index.js
noahmalmed: Isn't the whole point of this pr: https://github.com/zooniverse/Panoptes-Front-End/pull/3894 that a user can select for their workflow to be displayed in the app? I'm pretty sure we only set a workflow to `mobile_friendly` if a user makes a decision to have their workflow appear in the app.

So we aren't displaying all projects that meet the criteria, we are only displaying all of the apps that have met the criteria and have been selected to appear in the app. In that sense the attribute name `mobile_friendly` is misleading

wgranger:
Merging after our convo and updated https://github.com/zooniverse/Panoptes-Front-End/pull/3894

PR #156, opened 2018-02-20
Some Fixes found during Release

noahmalmed:
We weren't committing the version number for when we update the app. Looking at the history of the Android releases, they are all either 1.0 or 1.1 even though there has been like 10 releases. 

So I thought it may be useful to start committing the version to github

Describe your changes.
- Add some missing icons to Xcode project
- Increment the iOS and Android Versions
- Fix a production bug from the react upgrade
# Review Checklist

- [x ] Does it work in Android and iOS?
- [ x] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ x] Are tests passing?


wgranger:
Is it also worth updating the `buildToolsVersion` in the build.gradle? I'm getting some warnings on the build that the version is deprecated.

wgranger:
Ok. I got this running, and in the process was able to update Android Studio to a working version based on the guidelines set in https://facebook.github.io/react-native/docs/getting-started.html for sdks. Merging.

PR #158, opened 2018-02-22
Issue_130 Update Backend when tutorial is complete

### LINKED ISSUE #130: Repeated project tutorials
The tutorial for Camera Catalogue opens almost every time I open the project. I assume the app isn't updating the user project preference to record that the tutorial has already been completed.

noahmalmed:
Fixes #130 .

Describe your changes.
It has been one of my goals since I joined to make the app follow Redux best practices and paradigms. Mainly I want to organize the reducers into multiple smaller reducers that handle different portions of state. This is how redux suggests that it should be done: https://redux.js.org/basics/reducers#splitting-reducers and personally I think it is a much more verbose and clear way to use redux.

So because I was going to be making changes in the user portion of state to fix this bug, I implemented this design pattern and refactored some of the code. The main change is that state is now split up into to portions a `user` portions and a `main` portion. The `main` portion of state functions like the app used, but the `user` properly follows the best practices design pattern.

Most of the changes in the PR are related to this.

- Correctly report tutorial completion (bug fix)
- Use correct redux debugging module for react-native. On `remote-redux-devtools` [github](https://github.com/zalmoxisus/remote-redux-devtools) they suggest to use `redux-devtools-extension` instead. I missed this when I added the library. Fortunately their syntax is exactly the same so it's not a huge change.
# Review Checklist

- [x] Are tests passing?


src/reducers/userReducer.js
```diff
@@ -0,0 +1,49 @@
...
+
+export default function user(state=InitialUserState, action) {
+    switch (action.type) {
+        case ActionConstants.SYNC_USER_STORE: {
+            store.save('@zooniverse:user', { state })
+            return state;
+        }
+        case ActionConstants.SET_USER_FROM_STORE: {
+            if ( action.storeData !== undefined ) {
+                return action.storeData;
```
wgranger: Perhaps I am missing something here. Don't you need to make a copy of the state to retain the current key/values? `Object.assign()`.etc? I would think you would overwrite other items you're keeping track of in a line like this.

noahmalmed: So this is the way it was working before the refactor. Basically this actions purpose is to replace the current state with the one in persistent storage

wgranger: Ok. I suppose this makes sense as the state is referenced in each of the switch cases. It's just a different approach from what we've been doing on some of the other frontends, so I'm just used to that approach.

wgranger:
LGTM. Thanks for this PR. The changes here make sense and improve the code's readability.

PR #159, opened 2018-03-01
157: Fix Android device image bug

### LINKED ISSUE #157: Zoom by pinching
Our project (Notes from Nature) would like to be able to zoom into an image by "pinching" on the screen (not just tapping). I believe that this was the previous functionality. 

It is currently not working on Android version 1.1

We tried enabling Pan and Zoom, but that didn't work.

noahmalmed:
Fixes #157 

This PR Fixes a few different things
1.  Fixes 157 - React native has some issues with loading images on older Android devices. Mainly, there are a few specific things you have to do to avoid these problem. Increase the heap size on Android and configure Image components to use `resizeMethod="resize"`
2. In the process of fixing this issue, I spent time reconfiguring how we calculate the size of the are we display the images. Before, we had a hardcoded size that we used and then manually calculated the aspect ratio and sized the image into that area. Fortunately, RN will just do that for use, so I removed that step to simplify the whole thing.
3. Refactor the project are in the store.

# Review Checklist

- [x] Are tests passing?


src/actions/classifier.js
```diff
@@ -86,40 +86,39 @@ export function saveThenStartNewClassification() {
...
-      'metadata.session': getState().main.session.id,
-      'metadata.finished_at': (new Date).toISOString(),
-      'metadata.viewport': { width: getState().main.device.width, height: getState().main.device.height},
-      'metadata.subject_dimensions.0': subjectDimensions
-    }
+    Image.getSize(subject.display.src, (naturalWidth, naturalHeight) => {
+      const subjectDimensions = {
+        naturalWidth,
+        naturalHeight,
+        clientWidth: getState().main.device.subjectDisplayWidth,
```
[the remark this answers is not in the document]
wgranger: In sum from our in person chat, Will need some more checks here. Older Android devices would skip the `Image.getSize()` function which would skip the classification update and save on line 109-110.

src/actions/index.js
```diff
@@ -278,9 +208,11 @@ export function updateInterestSubscription(interest, subscribed) {
   var NotificationSettings = NativeModules.NotificationSettings
   return () => {
     return new Promise((resolve) => {
-      NotificationSettings.setInterestSubscription(interest, subscribed).then((message) => {
-        return resolve(message)
-      })
+      if (subscribed !== undefined) {
```
wgranger: Cool. I think this will take care of one of the errors I've been experiencing in testing.

src/components/ZoomableImage.js
```diff
@@ -1,49 +1,29 @@
...
-      </ImageZoom>
-    )
-  }
+const ZoomableImage = (props) => {
+  return (
+    <ImageZoom
+      cropWidth={Dimensions.get('window').width}
+      cropHeight={Dimensions.get('window').height}
+      imageWidth={Dimensions.get('window').width}
+      imageHeight={Dimensions.get('window').height/2}
```
wgranger: Why is this divided?

src/actions/classifier.js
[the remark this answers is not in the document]
noahmalmed: Added some error handling. Tested on the Android device and everything looks good there.

src/components/ZoomableImage.js
noahmalmed: I put this in just because it looked right, but forgot the change it. Using the device height value we use for all of the other images

noahmalmed:
@wgranger this ready to be reviewed again

wgranger:
Cool. LGTM. Good call using the optional failure argument on the `Image.getSize()` function.

Issue #144, opened 2018-02-06
Issue with planetary response network


noahmalmed:
Going to this page breaks the header.

noahmalmed:
Waiting on design decision from @beckyrother

beckyrother:
@noahmalmed the best solution would be to truncate the name of the project to the number of characters that comfortably fit in the header + `...`, so `planetary response ne...` or whatever fits. 

it would be great to do this with every project, including native mobile projects, so that navigation stays consistent.

PR #162, opened 2018-03-09
144: Fix Header

noahmalmed:
This is a bug fix for issue #144. Basically whenever a header title was long it would overlap the other header buttons and basically trap the user on the page because the text would block the buttons.

I refactored the nav bar code to use relative positioning instead of absolute positioning and I noticed that all of the pages on the app had a `topPadding` set to the size of the header. So that is why I have touched so many files in this PR.
# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/About.js
+++ b/src/components/About.js
@@ -71,7 +71,6 @@ class About extends React.Component {
 const styles = EStyleSheet.create({
   container: {
     flex: 1,
-    paddingTop: 70,
   },
   header: {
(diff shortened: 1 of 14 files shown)


wgranger:
Cool. This is a good fix. I noticed there were some absolute styling issues with #134, and it's good you picked it up and got a fix in. This will help changing any position styling in the future.

This branch looks good on Android and iOS (including iPhone SE). Merging.

PR #165, opened 2018-03-21
Fix image cropping

noahmalmed:
Fixing some issues caused by  #159.

Namely that I was not correctly implementing the automatic resizing provided in react. Before this code change, the images were being cropped to be a specific size, now they are actually scaled down.

Unfortunately, we still have to use `Image.getSize` to calculate the size of the overlay view. However, now if that function fails, we still see the image, the overlay view will just be a little to big.

We have to use this function because React Native doesn't provide any way to know the size of an automatically resized image.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/SwipeSubject.js
+++ b/src/components/SwipeSubject.js
@@ -3,12 +3,14 @@ import {
   Animated,
   Easing,
   Image,
+  ImageBackground,
   View
 } from 'react-native';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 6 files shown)


noahmalmed:
@wgranger I have a user waiting on this fix. I don't want to rush you too much, but if you could get this reviewed sooner than later that would be great.

wgranger:
Cool. This looks good.

Issue #161, opened 2018-03-07
Create Test Section in app for project builders


noahmalmed:
Inspired from #157 and from other talks. It would be very useful for project builders to be able to see their projects in the app before they actually release them to the public.

PR #166, opened 2018-03-28
Beta Project Viewer(Part 1)

noahmalmed:
# Description 
Completing the first part of #161

This PR includes every part of the workflow up to the actual swipe classifier, which will come in a follow up PR. 

# Details
- Add code to change nav bar color
- Add code to load beta projects
- Add code to display beta projects on discipline page and project page
- Update some design
- Update unit tests

# Review Checklist

- [x] Are tests passing?


wgranger:
I'm getting some failing tests. Make sure your client is set for the correct environment (instead of switching to 'production' on the client module). There are 11 tests that fail.

noahmalmed:
Hmm I'm not seeing anything after doing a delete node_modules and a reinstall.

I'm guessing something is wrong with your setup because CI tests are also passing.

noahmalmed:
@wgranger this should do the trick

src/components/NavBar.js
```diff
@@ -164,7 +165,14 @@ const styles = EStyleSheet.create({
   },
   disabledIcon: {
     color: '$transparent'
-  }
+  },
+  totalClassifications: {
+    color: '$darkTextColor',
+  },
+  userName: {
```
wgranger: Is this, and the style above, being used in this component?

src/components/ProjectDisciplines.js
```diff
@@ -28,13 +28,15 @@ const mapStateToProps = (state) => ({
   isGuestUser: state.user.isGuestUser,
   isConnected: state.main.isConnected,
   isFetching: state.main.isFetching,
-  pushPrompted: state.user.pushPrompted
+  pushPrompted: state.user.pushPrompted,
```
wgranger: Not sure this is being used. Instead, line 64 uses `this.props.user.pushPrompted`, which seems redundant.

src/components/projects/ProjectList.js
```diff
@@ -73,22 +85,36 @@ class ProjectList extends Component {
...
-        if (this.props.swipeEnabledProjects.length > 0) {
-            sections.push({data: this.props.swipeEnabledProjects, title: 'Made For Mobile'});
-        }
-
-        if (this.props.nonSwipeEnabledProjects.length > 0) {
-            sections.push({data: this.props.nonSwipeEnabledProjects, title: 'In-Browser Experience'})
+        if (inTestMode) {
+            if (!R.isEmpty(ownerIds)) {
+                const ownerProjects = swipeEnabledProjects.filter((project) => ownerIds.includes(project.id))
+                sections.push({data: ownerProjects, title: 'You Projects'})
```
wgranger: Your projects?

wgranger:
screen shot 2018-03-29 at 12 07 19 pm

This is the view on an iPhone SE. Notice how the text wraps onto the second line and looks a bit odd. I can't find exactly in this PR where this difference would've been made, but it isn't happening on an SE with the current version of the app.

wgranger:
Nice. Looks good.

PR #169, opened 2018-04-09
Beta Project Viewer(Part 2) 

noahmalmed:
### Overview 
Second half of #166. Adding functionality for a test classifier. Basically its just a normal classifier in terms of the UI, but none of the classifications get reported.

In addition, I did some heavy refactoring in the classifier code. 

### Details
 - Refactoring
   - Create an area in state that handles that classifier data and remove it from the legacy main data line.
   - Remove extraneous service calls that were called for every classification
   - Parallelize initialization code to quicken the initialization of classifiers

# Review Checklist

- [x] Are tests passing?


src/components/projects/ProjectTile.js
```diff
@@ -63,14 +97,24 @@ class ProjectTile extends Component {
...
+            this._navigateToSwipeClassifier(R.head(workflows))
         } else if (redirect) {
             this._openURL(redirect)
         } else {
             Actions.ZooWebView({project: this.props.project})
         }
     }
 
+    _navigateToSwipeClassifier(workflow) {
+        this.props.classifierActions.clearClassifierData()
```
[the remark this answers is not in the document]
noahmalmed: fixed

src/components/SwipeClassifier.js
```diff
@@ -200,9 +186,8 @@ export class SwipeClassifier extends React.Component {
         this.props.needsTutorial ? tutorial : classificationPanel
       )
     }
-
     return (
-      this.props.isFetching || isEmpty(this.props.workflow) ? <OverlaySpinner overrideVisibility={this.props.isFetching} /> : renderClassifierOrTutorial()
+      this.props.isFetching || !this.props.isSuccess ? <OverlaySpinner overrideVisibility={this.props.isFetching} /> : renderClassifierOrTutorial()
```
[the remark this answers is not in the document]
noahmalmed: fixed

src/actions/classifier.js
```diff
@@ -55,120 +54,85 @@ export function startNewClassification(workflowID, firstClassification=true) {
...
-    dispatch(removeState(`classifier.annotations.${workflowID}.${task}`, value))
-  }
-}
+    const classifier = getState().classifier
+    const classification = classifier.classification[workflow.id]
+    const subject = classifier.subject[workflow.id]
+    const structureAnnotation = (a) => {
+      return {task: a[0], value: a[1] } 
+    }
+    const annotations = R.map(structureAnnotation, R.toPairs(classifier.annotations[workflow.id]))
```
wgranger: This seems very odd to me, yet I know it is the code that was previously implemented. Can you make sense of it, or do you see it causing an issue in the future if we bring other workflows into the project instead of the swipe classifier?

src/reducers/index.js
```diff
@@ -79,5 +63,6 @@ export default combineReducers({
   main,
   user,
   projects,
-  navBar
+  navBar,
+  classifier
```
wgranger: Per convo w/ Noah. This is the line that is responsible for removing classifier out of the sort of `getState().main.classifier` (temp solution) and into the more organized `getState().classifier`.

wgranger:
This code looks pretty good. There are many changes, but many of them appear to be for the sake of organization. I did experience one bug when viewing the beta project "is it red?..." while trying to access the "new workflow" workflow.

src/actions/classifier.js
noahmalmed: I imagine this will be fine if we start accepting other workflows. `classifier.annotations` is normally in the following format: `{ t0: 1 }` or something like that. So if we, for example, allow for more answers then it will look like `{t0: 3}`. and it will still work fine.

noahmalmed:
@wgranger fixed the bug. I wasn't checking for whether workflows were swipe_verified which is just the client side check for wether we can show them or not. Also, I'm pretty sure this project was added from our hack day pfe work. So I think it's sort of in a weird state (marked as swipe_friendly, when it's actually not)

noahmalmed:
Just want to add some copy changes, before final review on this

noahmalmed:
@wgranger please review. All comments and requested changes implemented

wgranger:
Looks like some of the snapshots need to be updated.

noahmalmed:
My bad

wgranger:
Nice. This PR fixes #173 with the most recent commit. I think the preview section is shaping up nicely.

PR #177, opened 2018-04-25
Fix Out of Data Bug

noahmalmed:
Describe your changes:

I was checking for out of data incorrectly. Instead of checking the workflow completeness I was checking the project completeness. 

Also, added out of data checks for projects with multiple workflows.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/projects/ProjectTile.js
+++ b/src/components/projects/ProjectTile.js
@@ -22,8 +22,10 @@ import Separator from '../common/Separator'
 import PopupMessage from './PopupMessage'
 import theme from '../../theme'
 
+const horizontalPadding = 25
+
 const mapStateToProps = (state, ownProps) => ({
-    tileWidth: state.main.device.width - 52,
+    tileWidth: state.main.device.width - 2*horizontalPadding,
     containsNativeWorkflows: ownProps.project.workflows.length > 0,
 });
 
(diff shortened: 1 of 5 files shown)


noahmalmed:
@wgranger Can you take a look at this? It's been open for a while and I want it to get in before I make a release of the app. Thanks!

src/components/projects/ProjectTile.js
```diff
@@ -22,14 +22,10 @@ import Separator from '../common/Separator'
...
-const mapStateToProps = (state, ownProps) => {
-    const completeness = Number.parseFloat(ownProps.project.completeness)
-    return {
-        tileWidth: state.main.device.width - 52,
-        containsNativeWorkflows: ownProps.project.workflows.length > 0,
-        outOfData: !Number.isNaN(completeness) && completeness >= 1
-    }
-};
+const mapStateToProps = (state, ownProps) => ({
+    tileWidth: state.main.device.width - 52,
```
wgranger: What does 52 represent here? Might be good to have as a constant so it is more apparent.

wgranger:
This code is nice and readable, but there is just the one change requested with the constant. I also noticed the bug of a project being listed multiple times (below). This might not be blocking, however, as it appeared in the past and didn't leak over into the release.
screen shot 2018-05-08 at 8 55 59 am

PR #180, opened 2018-05-08
Update Version Strings

noahmalmed:
Describe your changes.
Just updating the version strings for the release.

# Review Checklist

- [x] Are tests passing?


wgranger:
Good catch. Sorry for the late review.

PR #185, opened 2018-05-23
Redesign Swipe Classifier

noahmalmed:
### Description:
UI design updates to the swipe classifier.

Doing a few different things here:
- Pulling a swiper library to handle the swipe code. I actually made a zooniverse fork of it [here](https://github.com/zooniverse/react-native-deck-swiper) just to add some functionality that the library was missing.
- Modifying how the swiper works in state. Now we just have a list of subjects that we iterate through. Instead of relying on a system that updates for upcoming and current subject on every single swipe. This should cut down on the complexity of state management.
- Organize the classifier components into their own directory

# Review Checklist

- [x] Are tests passing?


src/components/classifier/SwipeClassifier.js
```diff
@@ -0,0 +1,313 @@
...
+import * as classifierActions from '../../actions/classifier'
+import * as navBarActions from '../../actions/navBar'
+import Theme from '../../theme'
+import SwipeCard from './SwipeCard'
+import TaskHelpModal from './TaskHelpModal'
+import FontedText from '../common/FontedText'
+import { getTaskFromWorkflow, getAnswersFromWorkflow } from '../../utils/workflow-utils'
+
+export const subjectDisplayHeight = Dimensions.get('window').height - 380
+export const subjectDisplayWidth = Dimensions.get('window').width - 80
```
wgranger: Could we set these numbers (380 and 80) to constants to know what they signify?

wgranger:
Wow, this code makes a lot of changes and cuts many lines. It does look well organized. Overall, things look good, but I'm curious about project that are out of data.

- Is this a recent change that we can't access projects that are out of data? In the past, users have still been able to work on projects, even if they are marked out of data. Not sure when/why the app would perform differently, unless I missed a convo somewhere.

noahmalmed:
@wgranger Re: Projects being out of data. Oops, I put that in just to mess around to see how it felt. Guess I forgot to take it out

noahmalmed:
@wgranger updated per your review

wgranger:
Looks good. Merging.

Issue #179, opened 2018-04-30
Show some sort of loading feedback when images are loading


noahmalmed:
We currently don't give the user any sort of feedback when a subject is loading.

This can lead to two bad scenarios:
1) Neither image load and the user just sees a large patch of white space on their screen.

2) The back image loads before the front and the user will try to classify the back image.

These scenarios are both pretty bad (especially number two). I think we need some sort of feedback to the user that images are loading.

Maybe just be as simple as putting a placeholder image where the subject image should be.

@beckyrother any ideas?

beckyrother:
Agreed @noahmalmed . I think we have a loading spinner that's used elsewhere – would that be appropriate here as well?

noahmalmed:
![imagesloading](https://user-images.githubusercontent.com/10929131/39769553-25667466-52b2-11e8-818e-2c42bb042ea8.gif)

Here's a gif of what images loading often looks like in the app. The blank screen is when both images are loading, the next screen is when the back image has loaded and the front hadn't loaded yet, the last screen is both images loaded. 

I think putting up a loading spinner over the screen while both images load isn't a great solution because it may be disrupting.

Maybe some sort of placeholder gif or animation, even a spinner, where the image is loading would be enough.

noahmalmed:
![loading_example](https://user-images.githubusercontent.com/10929131/40685848-f77bd4e2-635a-11e8-845e-1254a996dc72.gif)

@beckyrother Here's a prototype for

beckyrother:
This looks great! Just use Karla for the loading text so it looks the same as that first spinner.

PR #187, opened 2018-06-06
Add Loading indicator for Classifier images

noahmalmed:
The issue for reference #179.

### Description

Unfortunately react-native doesn't really have an api for showing a placeholder view before an image loads. To make up for this, we have to download the image, and once it is finished downloading display the image.

We achieve this by first downloading the image to the devices cache and then passing it to the `ProgressIndicatingImage` class which will display a spinner until it is given a valid image.

Having the image downloaded also gives us some networking benefits. We no longer have to use the react-native `Image.getSize` API which led to us sometimes download the same image twice. We instead achieve this by building using the native library I built that natively calculates the image dimensions of an image on file.

We also don't have to re-download the image again when we display it in full screen mode.

# Review Checklist

- [x] Are tests passing?


src/components/classifier/SwipeCard.js
```diff
@@ -19,24 +31,57 @@ class SwipeCard extends Component {
...
+                return
+            }
+
+            const nativeImage = new NativeImage(localUri)
+            nativeImage.getImageSize().then(this.updateOverlayImageForImageDimensions)
+
+            this.props.imageActions.saveImageLocation(remoteUri, localUri)
             this.setState({
-                overlayAnswerIndex: value.value < 0 ? 0 : 1
+                localUri: localUri
```
wgranger: This can be `this.setState({ localUri })`, but I wouldn't worry about it unless there's anything else that I find needs adjusting here.

noahmalmed: I'll update

wgranger:
Code looks good, but this error occurs when I make a classification on the staging "Is this red?" project.

screen shot 2018-06-08 at 10 28 57 am

wgranger:
Also, this error is found on iOS whenever an image is begun swiping one way or another. I think you need to give unique keys to children in that component.

screen shot 2018-06-08 at 10 32 19 am

noahmalmed:
@wgranger can you double check that you've cleared the node modules folder and restarted the bundler. I'm not seeing either of these error on my machine

noahmalmed:
The bundler is really finicky and things often right themselves with a
refresh of the app.

Good news is that we don't use the bundler for release builds. That being
said, I'll keep my eyes out for these errors when I make a release build

wgranger:
Looks like the issues experienced here are due to the project being tested only having one image in a subject set. Issue #189 has been opened in lieu of a fix here.

wgranger:
This will need a rebase.

PR #195, opened 2018-07-17
Beta Part1: Add Forms Support

noahmalmed:
I'm trying to break this work up into small chunks, so this is just a portion of all of the Beta work.

Basically all I'm doing here is allowing the classifier to be configured for beta mode. 

When the classifier is in Beta mode, there is a small bottom bar that displays a message about beta testing and when you tap it will take you to a google form.

Also, just a note. This PR contains pending changes from #194. Should probably hold off on reviewing this until that other PR is merged

# Review Checklist

- [x] Are tests passing?


src/utils/googleFormUtils.js
```diff
@@ -0,0 +1,9 @@
+export const filledInFormUrl = (projectName, projectId, phoneType) => {
+    const baseUrl = 'https://docs.google.com/forms/d/e/1FAIpQLScq85a1bh9QRgSsKpUOfXB0B-o1y5llJx3AyWhxGyq--LtCrw/viewform?usp=pp_url&'
```
wgranger: Can you explain how this form works? It seems a bit odd to have this base url hardcoded.

wgranger:
It would be nice if some fields in the Google form (eg. ProjectID) could be made non-editable, so that they must retain the auto-filled in value. However, if this isn't possible, it shouldn't block this PR.

src/utils/googleFormUtils.js
noahmalmed: Explained in person. Basically we only have one form for all projects and we autofill some of the fields based on which project is being reviewed

noahmalmed:
Unfortunately I couldn't find anything having to do with disabling user input on a question.

wgranger:
Sounds good. Thanks for checking. I think this PR should be good to go then.

src/utils/googleFormUtils.js
srallen: You could POST directly to a Google Sheet if the form and responses are all shared anyway and write your own form with hidden auto filled inputs.

srallen: I just mean write your own web form with the inputs for the questions being asked and the `onSubmit` event handler would POST to the Google Sheets API directly rather than sending them over to a Google Form with auto-filled inputs that are vulnerable to editing.

Sheets API append method that takes a spreadsheet ID as a parameter: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/append

Hidden input types: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden

I've done a lot of Google API work in the past couple of years and we almost went this route for a Intro 2 Astro feature.

srallen: Or the update method might be a better option to submit to a Google Sheet: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/update

I forgot to say, Google Forms, just submit to an associated Google Sheet, so if Google Forms doesn't do what we want it to be doing, we could write our own web form instead.

noahmalmed: Oh for sure. Yeah, ideally we just make a native form for the user to fill out, but the whole purpose of using google forms is to just have a quick solution that anyone on the team can go in and modify. I think Google forms gives us most of the important functionality with almost no work.

PR #200, opened 2018-08-08
Correct Android Push Notification Icon

noahmalmed:
There was an issue where the Android Push Notification Icon would not show up correctly on later versions of Android. This code fixes the issue.

# Review Checklist

- [x] Are tests passing?


wgranger:
I assume this will only appear if I run a push notification. Do you recall where the documentation is located for creating a push notification. I assume I'll have to run it on a physical device.

noahmalmed:
Re documentation: Check the mobile github Readme 😏 

This is for Android only, so you can use the test phone I have

wgranger:
Looks great. Notification icon appeared in Android.

PR #201, opened 2018-08-16
Drawing Tool Part1: Initial work

noahmalmed:
**NOTE**
Please note that I made a new branch for these changes on the zooniverse repo, so all the app does now is go to the drawing screen

This is the first part of the drawing tool. Even though the code isn't complete, I wanted to put up a PR so you can take a look and give some feedback on the direction I'm heading in.

So basically this code displays (for the most part) what the user will see when they begin doing the actual drawing task. So the user should be able to draw squares, modify their sizes and delete the squares they draw.

Most of the classes should have descriptions in either the code or the commit message

# Review Checklist

- [x] Are tests passing?


wgranger:
Ok. I'll run through this and look over the code. Can you look into this error? I'm unable to test this branch.
screen shot 2018-08-17 at 9 52 47 am

src/components/Markings/components/DragableSquare.js
```diff
@@ -0,0 +1,241 @@
...
+                this.setState({
+                    x: this.props.x + dx,
+                    y: this.props.y + dy,
+                    width: this.props.width - dx,
+                    height: this.props.height - dy
+                })
+            },
+            (evt, gestureState) => {
+                const { dx, dy } = gestureState;
+                this.props.drawingScreenActions.mutateShapeAtIndex({
```
wgranger: This feels a bit odd that the props are mutated here, as the state is updated above, I'm assuming on the `onPanResponderMove`. What is happening here with `mutateShapeAtIndex`?

src/components/Markings/components/SvgOverlay.js
```diff
@@ -0,0 +1,177 @@
...
+
+    render() {
+        return (
+            <Svg
+                { ...this.panResponder.panHandlers }
+                style={styles.svg}
+                preserveAspectRatio="xMidYMid meet"
+            >
+                { this.state.isDrawing ? 
+                    <Rect
```
wgranger: Will this module likely need some rewriting in the future as more shapes are introduced? I noticed in the `DrawingTitle` component you have a case to determine what the shape is, where rect is currently the only option.

wgranger:
Does this branch delete shapes that are too small? In PFE, there is a minimum shape size requirement. If the rectangle is under the minimum size, the square is deleted. This is to keep users from accidentally tapping on the screen or creating tiny, dot-sized rectangles on accident that clutter up the UI.

It didn't look like there was code to account for this in the branch, but I'm unable to get this branch running, so I might've missed it.

src/components/Markings/components/DragableSquare.js
noahmalmed: So this is sort of what I'm trying to explain in the comments on line 50. 

Unfortunately, because react-native doesn't support animating SVGs, you need to either do animations by updating the state or props of components with the pan responder callbacks or by modifying components props directly using `setNativeProps`. Both are fairly hacky and each have their pros and cons.

Updating props and state: Is fairly verbose and the way you are "supposed to" update components, however it's performance can suffer when a lot of updates are happening.

Using `setNativeProps`: More performant, much harder to keep track of state. I had played around with this and was having a lot of trouble getting it to work well.

So I elected to use updating props and state, but I still wanted to cut down on the number of re-renders I do. So I used a component localized state for when he user is dragging the component to mutate it and once the gesture ends then we update global state (calling `mutateShapeAtIndex`) which updates all the other components that care about it (`DrawingScreen`).

To sum up, I use `setState` for drag animations and once the animation is complete , I update redux state on how the shape was mutated. I do all this to save on performance

src/components/Markings/components/SvgOverlay.js
noahmalmed: Yeah, I'll add something here to make it more modular

eatyourgreens:
I’d really like to change PFE so that drawing starts with a long press, to avoid the accidental tap problem. That’s how Operation War Diary was set up to work on iPad. OWD had a delay between the initial touch and creating the new mark, and only created the mark if you were still touching the same point at the end of the delay.

wgranger:
Issue resolved. I was running npm v.5.0.3 as opposed to 5.3.0. Changing versions resolved the dependency issue. 

Current Issue I'm seeing is:
screen shot 2018-08-20 at 9 42 44 am

wgranger:
Jim's comment makes sense, if introduced in this PR or another. It might be that users wouldn't need any explanation to do this since users are probably/hopefully keeping their finger on the screen longer than 0.5-1 second if they're making a mark.

My only other comments (after having tested this branch on the simulator) concerns resizing the square: 

- When resizing to the right, the square can accidentally grab the drawer, causing that to have drag priority.
- When resizing vertically, the border can go off screen, making it easy for the drag handlers to be lost off the image.

noahmalmed:
@eatyourgreens Long press is an interesting idea and I think we can do some testing on what feel right. But I'm not 100% sure that we need it a because in the app the user will have different modes they can toggle between (drawing mode, edit mode, and erase mode). This way it is a lot more explicit when the user is trying to draw or trying to tap the screen for other reasons

noahmalmed:
@wgranger changes made per your notes

src/components/Markings/components/DragableSquare.js
wgranger: I think that's a good solution. If the code hasn't changed, PFE calls `setState` whenever the shape changes at all, whether on drag or resize, and I always thought there might be a better solution to this. I think it is a good goal to cut down on the number of rerenders taking place.

wgranger:
These changes look good. I know this is a work in progress, but it looks like it's being merged into a `drawingTool` branch (as opposed to master). That said, is this good to merge if all looks well?

noahmalmed:
Yep! I have some more changes coming soon

PR #202, opened 2018-08-30
Update Android Manifest

noahmalmed:
Fixes a bug where two icons show on Android.

Testing:
Installing the Android app on a simulator or device will only show one icon. 

# Review Checklist

- [x] Are tests passing?


wgranger:
Looks good

PR #204, opened 2018-09-12
Add Fastlane

noahmalmed:
This pull request adds [Fastlane](fastlane.tools) to the project! Fastlane is a tool written in Ruby that helps you write scripts for building Android and iOS.

Most of the changes are detailed in the README update.

This is just a first pass at integrating fastlane, so for now all it does is build Android and iOS and push the builds to the Google Play Beta and Testflight.

In the future, we can integrate this into our CI process for building and deploying to a beta service

# Review Checklist

- [x] Are tests passing?


fastlane/VersionUtils.rb
```diff
@@ -0,0 +1,35 @@
...
+  def VersionUtils.versionStringIsCorrectLength (versionString)
+    versionStringArray = versionString.split(".")
+    return versionStringArray.length == 3
+  end
+  
+  def VersionUtils.isNewVersionBiggerThanOldVersion (newVersionNumberString, oldVersionNumberString)
+    newVersionArray = newVersionNumberString.split(".").map { |numberString| numberString.to_i }
+    oldVersionArray = oldVersionNumberString.split(".").map { |numberString| numberString.to_i }
+  
+    for index in 0..2
```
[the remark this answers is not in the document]
noahmalmed: ignoring this because we need the actual index

fastlane/version_utils.rb
```diff
@@ -0,0 +1,29 @@
+module VersionUtils
```
noahmalmed: @zwolf feel free to review the whole PR. But this is the only Ruby file (besides the fastfiles which I think are Ruby derivative

README.md
```diff
@@ -18,9 +18,11 @@ You'll need at least one emulator.  I've found the easiest way to add Virtual De
...
   *  Create at least one using the latest Android Release (Nougat).  I have a few different size and Android Release configurations
 
-Also you'll need a Gradle properties file outside the project for keeping secrets that aren't checked into source control.  This should be in ~/.gradle/gradle.properties and contain the following:
-`MYAPP_RELEASE_STORE_FILE=
-MYAPP_RELEASE_KEY_ALIAS=`
+Also you'll need a Gradle properties file outside the project for keeping secrets that aren't checked into source control. This should be in ~/.gradle/gradle.properties and contain the following:
+`MYAPP_RELEASE_STORE_FILE=`    
+`MYAPP_RELEASE_KEY_ALIAS=`
+
+Which should contain the keystore file name and alias. Talk to either Will Granger (will@zooniverse.org) or Noah Malmed (noah@zooniverse.org) for more information on this.
```
wgranger: We should think of other places to keep this keystore file instead of it being dependent on us.

README.md
```diff
@@ -40,20 +42,78 @@ react-native run-android
...
+- Install Fastlane using either  
+RubyGems: 
+	`sudo gem install fastlane -NV`  
+or Hombrew `brew cask install fastlane`
+
+
+#### iOS Preperations: 
+   Before getting everything setup it is probably worth spending some time understanding how [Fastlane match](https://codesigning.guide/) works. 
+
+- Make sure that you have access to the [private repo](https://github.com/zooniverse/mobile-provisioning-profiles) that houses our provisioning profiles.
```
wgranger: What is this? I don't think I have access.

android/app/build.gradle
```diff
@@ -103,8 +103,8 @@ android {
         applicationId "com.zooniversemobile"
         minSdkVersion 16
         targetSdkVersion 27
-        versionCode 17
-        versionName "2.1.0"
+        versionCode 23
+        versionName "2.2.7"
```
wgranger: Why such the increment? Was this all in getting Fastlane setup?

noahmalmed: Yeah, are version numbering is fairly arbitrary already. And I had to push a lot of builds up to the app store/play store consoles (production wasn't updated) for testing. So I figured I would just leave the versioning at what it ended up at at the end of testing

README.md
noahmalmed: There is a way to have google manage this for you, I'll write up an issue

README.md
noahmalmed: Check out https://codesigning.guide/ for an explanation. Basically this is a private repo that stores our provisioning profiles encrypted. And fastlane automatically pulls them down when you build so we only have to manage two provisioning profiles (developer and distribution) instead of two for each member of team. 

Basically this will make it easier in the future if we ever want to add a new member to the team or if any other developer wants to start making production builds

android/app/build.gradle
wgranger: That's fine. I recall the stores being picks about uping the version whenever a push is made.

README.md
noahmalmed: https://github.com/zooniverse/mobile/issues/205

noahmalmed:
@wgranger updated per your requesets

noahmalmed:
lets hold off on merge until @zwolf gets a chance to take a look

zwolf:
The version checking methods could be more simply written using Ruby's Gem::Version lib:

https://ruby-doc.org/stdlib-2.5.1/libdoc/rubygems/rdoc/Gem/Version.html

Incoming is a commit that uses that lib, fixes a num/number typo, and uses Ruby's implicit return to yield booleans from the last evaluated expression. Give it a test to make sure it does what you expect.

noahmalmed:
@zwolf thanks for the commit!

noahmalmed:
Inspired by @zwolf's changes, I updated the script to use a gem that handles version updating and removed my custom code. You now just pass in either 'major', 'minor' or 'patch' and the script will make the increment update for you

PR #206, opened 2018-09-28
Drawing Add Drawing Screen

noahmalmed:
Now that we have designs, I'm going to start committing to master instead of the other branch. 

I'm going to port over some of the stuff that we merged in there in future PRs.

# Changes

Detailed explanations for the changes are detailed in the commit messages.

- Added a navigator class that handles navigation to classifiers
- Refactor a bunch of components out of SwipeClassifier so they can be used in the drawing classifier.
- Refactor Image Loading logic
- Add DrawingScreen

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/Markings/DrawingClassifier.js
+++ b/src/components/Markings/DrawingClassifier.js
@@ -158,7 +158,7 @@ class DrawingClassifier extends Component {
                 >
                     {isQuestionVisible ? classification : tutorial}
                 </ClassificationPanel>
-                {isQuestionVisible && this.props.help && <NeedHelpButton onPress={() => this.classificationContainer.displayHelpModal()} /> }
+                {isQuestionVisible && !R.isEmpty(this.props.help) && <NeedHelpButton onPress={() => this.classificationContainer.displayHelpModal()} /> }
                 {isQuestionVisible && this.props.guide && fieldGuideButton}
                 { isQuestionVisible && submitButton }
             </View>
--- a/src/components/Markings/DrawingClassifier.js
+++ b/src/components/Markings/DrawingClassifier.js
@@ -1,0 +1,257 @@
+import React, { Component } from 'react'
+import EStyleSheet from 'react-native-extended-stylesheet'
+import {
+    View
+} from 'react-native'
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+import { bindActionCreators } from 'redux'
+import R from 'ramda'
+
+import Theme from '../../theme'
+import ClassificationPanel from '../classifier/ClassificationPanel'
+import ImageWithSvgOverlay from './ImageWithSvgOverlay'
+import Question from '../classifier/Question'
+import Tutorial from '../classifier/Tutorial'
+import * as imageActions from '../../actions/images'
+import * as classifierActions from '../../actions/classifier'
+import ClassificationContainer from '../classifier/ClassifierContainer'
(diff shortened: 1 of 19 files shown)


src/components/Markings/DrawingClassifier.js
```diff
@@ -0,0 +1,257 @@
...
+    finishTutorial() {
+        if (this.props.needsTutorial) {
+            this.props.classifierActions.setTutorialCompleted(this.props.workflow.id, this.props.project.id)
+        } else {
+            this.setQuestionVisibility(true)
+        }
+    }
+
+    render() {
+        if (this.props.isFetching || !this.props.isSuccess) {
```
wgranger: Is there still a way to display an error in the case `isFetching` is completed but `isSuccess` remains false, leaving the overlay spinning?

wgranger:
Just to confirm, this PR only contains the designs. There should not yet be any drawing functionality. Is that right?

src/components/classifier/ClassifierButton.js
```diff
@@ -0,0 +1,74 @@
...
+    TouchableOpacity
+} from 'react-native'
+import FontedText from '../common/FontedText'
+import PropTypes from 'prop-types'
+import EStyleSheet from 'react-native-extended-stylesheet'
+
+const ClassifierButton = (props) => {
+    const buttonStyle = [props.style, styles.button]
+    const textStyle = []
+    switch (props.type) {
```
wgranger: Will there be a default here, or are these the only types you are expecting that will expand in the future?

src/components/classifier/SwipeCard.js
```diff
@@ -124,7 +122,6 @@ class SwipeCard extends Component {
                     style={[styles.image, styles.imageShadow]}
                     resizeMethod="resize" 
                     resizeMode="contain"
-                    loadingText={'Loading Subject'}
```
wgranger: Is there any reason this was removed? Is this message now displayed to the user elsewhere?

src/components/classifier/SwipeClassifier.js
```diff
@@ -257,55 +240,30 @@ export class SwipeClassifier extends React.Component {
           { this.state.isQuestionVisible ? classification : tutorial }
           { this.state.isQuestionVisible ? unlinkedTask :null }
         </ClassificationPanel>
-        { this.state.isQuestionVisible ? needHelpText : null }
+        { this.state.isQuestionVisible && this.props.task.help ? <NeedHelpButton onPress={() => this.classifierContainer.displayHelpModal()} /> : null }
```
wgranger: Will this still evaluate to true if `this.props.task.help` is an empty object - `{}`? I think that is the default for a task, so you might want to double check this.

src/components/classifier/ClassifierButton.js
noahmalmed: This could expand in the future. I prefer to leave out the default so that if this component is being used incorrectly it will brake the code

src/components/classifier/SwipeCard.js
noahmalmed: Yes, that component actually doesn't take that prop. I think it got merged by accident.

src/components/classifier/SwipeClassifier.js
noahmalmed: good call. Will fix that

noahmalmed:
Component test will be incoming once the feature is complete because of the time crunch for the demo

PR #219, opened 2018-11-05
Drawing - Show Already Seen Banner, Clear in Progress Actions

### LINKED ISSUE #214: Ensure Already Seen Banner is Seen
The banner isn't reliably appearing with subjects seen the second time around during a session.

noahmalmed:
Fixes #214

Invision Mock-ups: https://projects.invisionapp.com/d/main#/projects/prototypes/13393648

# Changes
Couple of things here: 
	- Add an AlreadySeenBanner to the drawing Classifier if a user has already seen a subject.
	- Add a drawing action to clear just the users edits. Trigger this action when the user cancels the drawing.

# Review Checklist

- [x] Are tests passing?


src/components/Markings/components/ShapeEditorSvg.js
```diff
@@ -165,9 +164,14 @@ class ShapeEditorSvg extends Component {
...
                             showCorners={this.props.mode === 'edit' && selectedShape}
                             isDeletable={this.props.mode === 'erase'}
-                            ref={ref => this.setState({
-                                shapeRefs: R.set(R.lensProp(index), ref ,this.state.shapeRefs)
-                            })}
+                            ref={ref => {
+                                if (ref) {
+                                    this.setState(currentState => {
+                                        currentState.shapeRefs = R.set(R.lensProp(index), ref, currentState.shapeRefs)
+                                        return currentState
```
wgranger: What is the purpose of this change?

noahmalmed: There was a race condition in the code because this code was being called multiple times so close together it was essentially concurrent. React optimizes `setState` by batching updates and committing them at the same time. However, if you pass in a function to `setState` it will run the updates in sequence, thus preventing race conditions on the `state.shapeRefs` object.

Explained in: 
https://stackoverflow.com/questions/30338577/reactjs-concurrent-setstate-race-condition

wgranger:
Just one comment above. I can confirm that I see the "Already Seen" banner in the drawing workflow. I am also able to cancel out of annotations and only delete my most recent boxes.

I initially tried to view "Already Seen" in the "Is it red?..." test project, but instead I got to the end of the subject set and received the warning below. I believe this is an unrelated bug, but the expected behavior was to cycle through the subjects and eventually see the "Already Seen" banner on those repeated subjects. (Note: I was not logged in when doing this)

screen shot 2018-11-05 at 3 14 14 pm

noahmalmed:
Thanks, I'll take a look at that crash

noahmalmed:
@wgranger the crash should be fixed

wgranger:
LGTM

PR #220, opened 2018-11-08
Fix Tutorial View

### LINKED ISSUE #217: The tutorial view and modal's content is being cut off to the right

noahmalmed:
Fixes #217 

# Changes
Refactored how the tutorial is displayed. It is now displayed as a paginating scroll view. Appearance-wise it looks identical, but it fixes a visual issue where the image load lags behind the markdown load.

Using a paginating scrollview, all of the content loads at once. This means the transition is smooth between pages.

Also, fix a bug that caused a crash with image caching.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/classifier/Tutorial.js
+++ b/src/components/classifier/Tutorial.js
@@ -7,14 +7,14 @@ import {
   View
 } from 'react-native'
 import EStyleSheet from 'react-native-extended-stylesheet'
-import Markdown from 'react-native-simple-markdown'
 import Icon from 'react-native-vector-icons/FontAwesome'
+import Swiper from 'react-native-swiper';
 import { addIndex, length, map } from 'ramda'
 import PropTypes from 'prop-types';
 
 import StyledText from '../StyledText'
 import Button from '../Button'
-import FittedImage from '../common/FittedImage'
+import TutorialStep from './TutorialStep'
 
 const topPadding = (Platform.OS === 'ios') ? 10 : 0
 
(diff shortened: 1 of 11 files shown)


src/components/classifier/Tutorial.js
```diff
@@ -23,23 +23,27 @@ export class Tutorial extends Component {
...
 
   render() {
     const steps = this.props.tutorial.steps
-    const step = steps[this.state.step]
     const totalSteps = length(steps)
-
-    const mediaResource = (step.media ? this.props.tutorial.mediaResources[step.media] : null )
-    const mediaImage = (mediaResource !== null ? <SizedImage source={{ uri: mediaResource.src }} /> : null)
+    
+    const tutorialStep = steps.map((step, index) => {
```
wgranger: This should probably be `tutorialSteps` plural

src/components/classifier/Tutorial.js
```diff
@@ -23,23 +23,27 @@ export class Tutorial extends Component {
...
+      return (
+        <TutorialStep
+          key={`${index}`}
```
wgranger: Might be safer here to make the key `TUTORIAL_STEP_${index}` or something similar to avoid the key from interfering with any other z-indexed keys in the view.

noahmalmed:
@wgranger updated per your review

src/components/classifier/Tutorial.js
```diff
@@ -69,7 +69,7 @@ export class Tutorial extends Component {
     const renderCircle = (currentStep, index) => {
       return (
         <TouchableOpacity
-          key={ index }
+          key={ `PAGINATION_DOT_${index}` }
```
wgranger: Good catch here. Didn't see this the first time around.

wgranger:
Looks good. Thanks for the update.

PR #221, opened 2018-11-14
Add tablet specific code/Assorted refactors

noahmalmed:
# Description

There are a few things going on here:

1) Dynamicaly Size Text for Android
	- Add `react-native-device-info` this package lets us check if the device is a tabler
	- Add Dynamic sizing for text in the classifier and tutorial

2) Remove drawing modal for tablet
	- We removed the modal because we don't need to create additional space on tablet.

3) Various layout and formatting fixes
	- Fixed an issue with tutorial margins
	- Fixed a bug where the field guide always show on the drawing task

# Review checklist

- [x] Are tests passing?

### DIFF
--- a/src/components/Markings/DrawingClassifier.js
+++ b/src/components/Markings/DrawingClassifier.js
@@ -220,7 +220,7 @@ class DrawingClassifier extends Component {
                     {isQuestionVisible ? classification : tutorial}
                 </ClassificationPanel>
                 {isQuestionVisible && !R.isEmpty(this.props.help) && <NeedHelpButton onPress={() => this.classificationContainer.displayHelpModal()} /> }
-                {isQuestionVisible && this.props.guide && fieldGuideButton}
+                {isQuestionVisible && !R.empty(this.props.guide) && fieldGuideButton}
                 { isQuestionVisible && submitButton }
             </View>
 
--- a/src/components/Markings/DrawingClassifier.js
+++ b/src/components/Markings/DrawingClassifier.js
@@ -147,7 +147,7 @@ class DrawingClassifier extends Component {
 
         // We validate that tools has at least one element earlier
         const tool = this.props.tools[0]
-        const warnForRequirements = this.state.modalHasBeenClosedOnce && R.keys(this.props.shapes).length < this.props.tools[0].min
+        const warnForRequirements = this.state.modalHasBeenClosedOnce && R.keys(this.props.shapes).length < tool.min
 
         const tutorial =
             <Tutorial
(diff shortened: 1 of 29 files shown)


src/components/Markings/DrawingClassifier.js
```diff
@@ -141,6 +145,8 @@ class DrawingClassifier extends Component {
             return <OverlaySpinner overrideVisibility={this.props.isFetching} />
         }
 
+        // We validate that tools has at least one element earlier
+        const tool = this.props.tools[0]
```
wgranger: Is there a chance this could be null? (if used as validation, as the comment says) Also, `this.props.tools[0]` is still used in the line below, as opposed to this new variable.

src/components/Markings/DrawingClassifierSubject.js
```diff
@@ -113,36 +90,24 @@ class ImageWithSvgOverlay extends Component {
     }
 
     render() {
-        const pathPrefix = Platform.OS === 'android' ? 'file://' : ''
```
wgranger: Where has this, and other items for this component, been moved to? It seems this component has changed significantly.

src/components/Markings/components/DrawingButtons.js
```diff
@@ -12,25 +12,25 @@ const DrawingButtons = (props) => {
                 disabled={!props.canUndo}
                 type="undo"
                 activated={false}
-                onPress={() => props.onButtonSelected('undo')}
+                onPress={() => props.onUndoButtonSelected()}
```
wgranger: Should 'undo' be removed here, or is this now the default action?

src/components/Markings/DrawingClassifier.js
noahmalmed: The comment is supposed to point out that tools has been validated to have `length > 0` when we validate the workflow after loading from the API

src/components/Markings/DrawingClassifierSubject.js
noahmalmed: So before this refactor, there were two components that essentially did the same thing: `ImageWithSvgOverlay` and `MarkableImage`. They both rendered SVGs over an image, but `MarkableImage` also allowed the user to draw SVGs over the image as well.

As a part of this refactor, I changed `ImageWithSvgOverlay` to `DrawingClassifierSubject`, who's purpose is to display the `DrawingToolView`: which contains the subject and all drawing and SVG logic. And display the blurview and already seen banner over the subject.

In addition, I modified `MarkableImage` to be drawing enabled or disabled, so it can essentially function exactly the same as `ImageWithSvgOverlay` used to function.

So, in short, the functionality wasn't really moved anywhere. It already existed in `MarkableImage` and I'm just leveraging that now

src/components/Markings/components/DrawingButtons.js
wgranger: Ahh yes, my bad. I didn't initially notice the function changed, only the argument.

noahmalmed:
This is ready for re-review

wgranger:
Nice. Looks good on iOS and Android.

PR #224, opened 2018-12-04
Add Landscape Mode for Tablet

noahmalmed:
### Changes:
- Update the app to be based completely on flex layout so the app can layout correctly even on page layout.
- Add an appState that updates the screen dimensions whenever there is a screen rotation.
- Fix an issue with the mardown library we use

# Review Checklist

- [x] Are tests passing?


wgranger:
Is there a way to trigger this in the simulator? I can't get the views to update to landscape when I rotate the display, which is how I'm assuming this should work.

src/constants/actions.js
```diff
@@ -66,4 +66,6 @@ export const CLEAR_SHAPES = 'CLEAR_SHAPES'
 export const SAVE_EDITS = 'SAVE_EDITS'
 export const MUTATE_SHAPE = 'MUTATE_SHAPE'
 export const UNDO_EDIT = 'UNDO_EDIT'
-export const CLEAR_SHAPES_IN_PROGRESS = 'CLEAR_SHAPES_IN_PROGRESS'
```
wgranger: Is looks like `CLEAR_SHAPES_IN_PROGRESS` is still being used elsewhere in the code. Should it be removed there?

src/containers/zooniverseApp.js
```diff
@@ -34,20 +36,30 @@ const mapDispatchToProps = (dispatch) => ({
   setNotificationPayload(value) {
     dispatch(setState('notificationPayload', value))
   },
-  setDimensions() {
-    dispatch(setDimensions())
-  },
```
wgranger: `setDimensions` can be removed as a defined propType.

noahmalmed:
> Is there a way to trigger this in the simulator? I can't get the views to update to landscape when I rotate the display, which is how I'm assuming this should work.

Because this PR makes changes to the native code, you may need to delete the app that's on your simulator and reinstall.

src/constants/actions.js
noahmalmed: Oops, not sure why this got deleted!

wgranger:
Yep. Deleting the app took care of the landscape switching. There's a slight bug with how the tutorial is rendered.

On viewing a project for the first time, the tutorial image and text is visible during each step of the tutorial. However, if you get into the project classifier and click on the "tutorial" tab, the image size grows and often pushes the text below view (you can scroll to view the text). Not sure why the image size would change from the initial render to the second time around.

wgranger:
Nice. Takes care of the view on the tutorial. Merging.

PR #226, opened 2018-12-12
Drawing/issue225

### LINKED ISSUE #225: Update drawing tools
> UI: toggling between Edit and Add was frustrating and not obvious. Dragged squares off to delete them (this did not actually delete them). Needs to reset to the Add tool each time.

There are a few updates to the drawing tool we want to implement based on user feedback:

- Remove the Edit button and let users edit their markings immediately. 
- Keep the trash button but also allow users to drag markings off the image to delete. Also, add visual feedback to drag to delete action like making the shape more transparent.
- Drag the edges of boxes to move them around.

noahmalmed:
Fixes #225 

# Changes:
There were a few feedback notes from user testing that are detailed in the linked issue above.

- Remove the edit button and have drawing mode and edit be the same thing
- Change shape dragging gestures to only trigger on the borders and the corners of the shapes.
- Add functionality to drag shape off the screen to delete

# Review Checklist

- [x] Are tests passing?


wgranger:
Haven't combed through the code line-by-line yet, but I am consistently getting this bug when I try to select a box to then move it off page. 
screen shot 2018-12-12 at 2 55 08 pm

noahmalmed:
@wgranger Fix in for the reported issue. Description is in the commit comment. I'm having a lot of trouble reproducing this bug on my end. So it's possible this fix doesn't fix the bug

src/components/Markings/components/ShapeEditorSvg.js
```diff
@@ -94,36 +106,90 @@ class ShapeEditorSvg extends Component {
...
+                const { shapeIndex, touchState, shapeRefs, isDrawing } = this.state
+                const { locationX, locationY } = evt.nativeEvent
+                const { dx, dy } = gestureState
                 if (shapeIndex >= 0) {
-                    const { dx, dy } = gestureState;
                     // Because Svgs don't have any way to animate, we have to update their props manually
                     const deltas = calculateShapeChanges(touchState, dx, dy, this.props.displayToNativeRatioX, this.props.displayToNativeRatioY)
-                    shapeRefs[shapeIndex].update(deltas);
+                    const newDimensions = shapeRefs[shapeIndex].update(deltas);
+                    const shapeIsOutOfBounds = isShapeOutOfBounds(newDimensions, {width: this.props.width *this.props.displayToNativeRatioX, height: this.props.height * this.props.displayToNativeRatioY}) && touchState.onlySquare
```
wgranger: What does `onlySquare` reference exactly? I'm sure it's necessary, but I'm curious why based on the naming. Not sure why it needs to be true to determine if out of bounds.

src/components/Markings/components/ShapeEditorSvg.js
```diff
@@ -38,14 +54,17 @@ class ShapeEditorSvg extends Component {
...
                 bottomLeft: false,
                 bottomRight: false,
                 upperLeft: false,
                 upperRight: false,
                 onlySquare: false
             },
-            shapeRefs: {}
+            shapeRefs: {},
+            isDrawing: false,
+            previewShapeDimensions: this.previewShapeInitialDimensions()
```
wgranger: This seems odd. Why would `this.state.previewShapeDimensions` ever be needed as opposed to simply using `this.previewShapeInitialDimensions()` (with a different name, if needed)

wgranger:
One more item to check. I'm noticing on Android that the box tends to "bounce" back when I pull the bounds outside the image area. This only happens on the initial box draw and not when I grab one of the corner handlers to resize.

src/components/Markings/components/ShapeEditorSvg.js
noahmalmed: `onlySquare` is describing how the user is touching the shape. It means the user is touching the shape but not touching any of the corners. Essentially it means the square shouldn't be resized, just dragged. 

The reason we check it here is because the delete action only happens on drag out of bounds, not resize out of bounds. 

Let me change around some of the naming to make this more clear.

src/components/Markings/components/ShapeEditorSvg.js
noahmalmed: `previewShapeDimensions` gets modified in other places in the code, we only use `previewShapeInitialDimensions` to set the initial value. The reason we use a function for it is because it is based on some props passed in and those props may change during the lifecycle of the component.

wgranger: Gotcha. Saw it is also used when resetting items, but that makes sense.

wgranger:
Nice! Looking sharp.

PR #228, opened 2018-12-28
Swipe cards/just swipe card

noahmalmed:
Adding a new design for the swipe cards, as well adding support for multi image subjects.

Invision Mock-ups: https://projects.invisionapp.com/d/main#/console/13393648/333632079/preview

Describe your changes.
- Update `SwipeCard` with new design per invision mock ups
- Add support in the app for multi-image subjects
- Add UI for displaying multi-image subjects  

# Review Checklist

- [x] Are tests passing?


src/components/classifier/SwipeCard.js
```diff
@@ -99,64 +120,116 @@ class SwipeCard extends Component {
...
-                <Animated.View style={[imageDimensionStyle, styles.overlayContainer, styles.overlayBackground, { opacity }]}>
-                    <FontedText style={styles.overlayText}>
-                        { this.props.answers[this.state.overlayAnswerIndex].label }
-                    </FontedText>
-                </Animated.View >
-            </View>
+            <Animated.View style={[styles.overlayContainer, styles.overlayBackground, { opacity }]}>
+                <FontedText style={styles.overlayText}>
+                    { this.props.answers[this.state.overlayAnswerIndex].label }
+                </FontedText>lop
```
wgranger: Should this "lop" be here after `</FontedText>`?

src/components/classifier/SwipeCard.js
```diff
@@ -99,64 +120,116 @@ class SwipeCard extends Component {
...
+            </Animated.View >
 
         const pathPrefix = Platform.OS === 'android' ? 'file://' : ''
+        const dimensionsStyle = {width: this.props.subjectDisplayWidth, height: (this.props.subjectDisplayHeight - 50)}
```
wgranger: Why isn't the height here the passed in `subjectDisplayHeight`? I'm sure there's a good reason, it just seems odd with height using the original prop.

src/components/classifier/SwipeCard.js
```diff
@@ -99,64 +120,116 @@ class SwipeCard extends Component {
...
+                            })}>
+                            <VerticalViewPager
+                                bounces={false}
+                                showsVerticalScrollIndicator={false}
+                                ref={ref => this.pager = ref}
+                                key={localUris.length}
+                                // Make this happen while scrolling
+                                onScroll={({nativeEvent}) => {
+                                    this.setState({
+                                        imageIndex: Math.round(nativeEvent.contentOffset.y/pagerDimensions.height)
```
wgranger: This seems a bit odd, is the index being set here based on size proportions?

src/components/classifier/VerticalPaginationBar.js
```diff
@@ -0,0 +1,84 @@
...
+                <SimpleLineIcons 
+                    style={[styles.arrowIcon, props.pageIndex === props.totalPages - 1 ? styles.iconGrey : null]}
+                    size={chevronSize}
+                    name="arrow-down"
+                />
+            </TouchableOpacity>
+        </View>
+    )
+}
+//textLayer.textColor = UIColor.black
```
wgranger: Should this be removed?

src/components/common/ProgressIndicatingImage.js
```diff
@@ -23,10 +23,20 @@ class ProgressIndicatingImage extends Component {
...
+            },
+            imageDimensions: {
+                width: 0,
+                height: 0
+            }
         }
         this.animateImageIfLoaded = this.animateImageIfLoaded.bind(this)
         this.unlinkImageOnLoad = false
+
+        this.animateImageIfLoaded()
```
wgranger: Is there a reason this is done here instead of within a lifecycle method?

src/components/classifier/SubjectOptionsBar.js
```diff
@@ -0,0 +1,54 @@
...
+} from 'react-native'
+import PropTypes from 'prop-types'
+
+import CircleIconButton from '../common/CircleIconButton'
+
+const SubjectOptionsBar = (props) => {
+    return (
+        <View>
+            <View style={styles.contentContainer}>
+                {/* TODO: Add this in a future PR
```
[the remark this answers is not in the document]
noahmalmed: This is for a favorite button. Honestly I'm just going to remove this TODO. No reason for it to be there

src/components/classifier/SwipeCard.js
noahmalmed: Woops! That shouldn't be there

src/components/classifier/SwipeCard.js
noahmalmed: Nope. The idea here is that the scroll view has 'pages' of content, meaning that each image in the scroll view takes up the entire viewable size of the scrollview (`pagerDimensions`). So, in order to calculate the pager index, we divide the offset by the size of pager. By rounding this number, we get the nearest page that scroll view is at.

src/components/classifier/VerticalPaginationBar.js
noahmalmed: yee

src/components/common/ProgressIndicatingImage.js
noahmalmed: No, this should probably go in `componentDidMount`

src/components/classifier/SwipeCard.js
noahmalmed: This was added in attempt to fix a bug, but it didn't. I'm going to remove the `-50` and open an issue for the bug. I have a fix for it, but it should go in a separate PR bc it's actually not related to this code.

noahmalmed:
This should be ready for rereview

wgranger:
I think this is good to go after the recent additions.

wgranger:
Well, actually the only issue seems to be with scrolling on Android. If you swipe up quickly on a multi-image subject, the viewer will momentarily go to the last image in the deck before scrolling back to the next image in line. 

I'd be fine with opening an issue for this or waiting if you want to add a fix here on another commit.

noahmalmed:
@wgranger I think it would be better to open an issue for this. I'm pretty sure it's a problem with the a dependency that I'm using, so I'll have to make a PR to that code to fix the problem

wgranger:
Cool. I can write up an issue now. Will merge this current iteration.

PR #242, opened 2019-03-04
Add safe area view

### LINKED ISSUE #235: Visual bug on iPhone X
When deep-linking to the app on iPhoneX we see the following visual bug.
simulator screen shot - iphone x - 2019-01-22 at 09 32 13

Notice the back button that iOS puts in the upper right of the view. 

I think all that is required for this is to wrap the highest level element in a [`SafeAreaView`](https://facebook.github.io/react-native/docs/safeareaview). This, however, is a pretty low priority issue because it will only appear on iPhone Xs when we send out push notifications which we don't do that often.

noahmalmed:
Fixes #235

Describe your changes.
 - Add `SafeAreaView` to account for iPhoneX native screen elements
 - Update how we keep track of navbar color so we can change the `SafeAreaView` color appropriately.
 - Some UI tweaks to go along with `SafeAreaView`
 
# Review Checklist

- [x] Are tests passing?


wgranger:
Rebase needed after the recent merge.

noahmalmed:
Spacing should be fixed on iOS. Before adding the `SafeAreaView` we were making up for that space with static padding for iOS only. So I just removed that

wgranger:
LGTM

PR #243, opened 2019-03-06
Just grab the file type, no parameters

noahmalmed:
Fixes bug noted [here](https://www.zooniverse.org/talk/739/910315?comment=1509585&page=1)

The changes are pretty simple: 
 The code was assuming that between the last period and the end of an image url would be the extension like so: 

`https://www.imagehost.image.jpeg`

However, the code fails if the image url has parameters.

`https://www.imagehost.image.jpeg?height=256;width=256`.

So the new code change just scrubs out the parameters

# Review Checklist

- [x] Are tests passing?


wgranger:
LGTM. Good catch if you're just trying to get that file extension.

PR #244, opened 2019-03-11
Update Version string/ Update Xcode

noahmalmed:
Just an update to the app versions numbers.

As well, did some testing for the latest version of xcode and everything seems to be working.

# Review Checklist

- [x] Are tests passing?


wgranger:
LGTM

PR #245, opened 2019-03-21
Multiple answer question

noahmalmed:
This PR adds multi-question workflows to the app.

Invision Mock-ups: https://projects.invisionapp.com/prototype/Multiple-choice-cjt4r1bi600o3oo01zttw62ng

# Review Checklist

- [x] Are tests passing?


src/components/classifier/QuestionClassifier.js
```diff
@@ -0,0 +1,323 @@
...
+                  }
+                </ScrollView>
+              :
+                renderTutorial()
+            }
+            <FullScreenImage
+              source={this.state.fullScreenImageSource}
+              isVisible={this.state.showFullSize}
+              handlePress={() => this.setState({ fullScreenQuestion: '', showFullSize: false })}
+              question={this.state.fullScreenQuestion}
```
wgranger: Can you destructure these state items as you've done with several other items at the beginning of the `render()` function?

src/components/classifier/SwipeClassifier.js
```diff
@@ -121,7 +120,7 @@ export class SwipeClassifier extends React.Component {
             onExpandButtonPressed={this.expandImage}
             subjectDisplayWidth={this.state.swiperDimensions.width}
             subjectDisplayHeight={this.state.swiperDimensions.height}
-            dimensionsLoaded={this.state.swiperDimensionsLoaded}
+            dimensionsLoaded={true}
```
wgranger: Is there a reason this was changed to always be true? Looks like this is the only component using `SwipeCard`, and there is still a conditional check for opacity in the component dependent on this prop.

wgranger:
Also, I know real estate has always been an issue with mobile, and the auto-scroll does after answer selection does help, but first view of the classifier hides the "Need some help" and "Field guide" buttons. Is there any chance of adding a visual indicator (arrow?) when we could scroll down further?

If so, this addition can be in a future branch.

Screen Shot 2019-03-25 at 4 13 15 PM

src/components/classifier/SwipeClassifier.js
noahmalmed: I'm going to clean up the `swiperDimensionsLoaded`, I don't think we need it.

src/components/classifier/TapableSubject.js
```diff
@@ -0,0 +1,177 @@
...
+
+        this.state = {
+            scrollViewOffset: new Animated.Value(0),
+            imageIndex: 0
+        }
+
+    }
+
+    componentDidMount() {
+        this.state.scrollViewOffset.addListener(({value}) => {
```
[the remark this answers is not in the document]
noahmalmed: It just keeps track of the page that the scroll view is scrolled to. This allows us to do ui stuff with the pagination dot indicators

noahmalmed:
We can flash the scroll indicators on page load. As well, when the user selects an answer the scroll view should scroll all the way to the bottom revealing the help and field guide buttons. Maybe another solution is to put the submit button below the help and field guide buttons. @beckyrother any thoughts?

noahmalmed:
@wgranger changes made per your review

wgranger:
Great. Code is working much better since the recent commits. Just a couple comments. 

This is what the multi-answer workflow looks like on my iPhone SE. Granted, the SE is the smallest screen size, but I could see people complaining about the scroll down to see more answers. I'm not sure the best solution, since there could be any number of answers (5, 6, 7+), but this adds more to the importance of a scroll indicator.  
Image from iOS

Also, is it intentional the next image is visible alongside each other in a multi-image subject? I could see this being helpful to know there are multiple subjects, but I want to make sure this is the intended display.
Screen Shot 2019-03-28 at 9 06 37 AM

noahmalmed:
@wgranger For your first point, I think it's best we that we live this as is for a few reasons. 

1) A scroll indicator does show up when the user scrolls, which is pretty generic design for mobile.
2) We are running short on my time here, and its probably best we get this is for a release relatively early next week.

As for your second comment, yes it is intended.

wgranger:
Alright. Aside from those comments, functionally this PR is good to go.

PR #256, opened 2019-06-04
Update README to resolve an inconsistency that Will noticed while set…

chelseatroy:
…ting up on his local machine

Fixes # .

Invision Mock-ups: 

Describe your changes.

# Review Checklist

- [ ] Are tests passing?


wgranger:
Nice. Good call remembering this.

Issue #259, opened 2019-06-06
General Items to Address After Upgrade


wgranger:
After our most recent RN upgrade, there are several low-priority items that need to be addressed on the app. 

- Tests are running rather slow, and it takes about a minute to run several tests
- We need to address the RN warnings of items that will soon be deprecated to avoid issues with a future RN migration
- The README can be updated to include `react-native run-ios` as a valid command to run the app.

lcjohnso:
Closed by #260 and follow-on RN updates

PR #260, opened 2019-06-10
Address Update Issues

chelseatroy:
This PR addresses the items in the following issue: https://github.com/zooniverse/mobile/issues/259

- [x] Tests are running rather slow, and it takes about a minute to run several tests <-- So this one I am struggling to repro. We might need to compare our machines or something to figure this one out.
- [ ] We need to address the RN warnings of items that will soon be deprecated to avoid issues with a future RN migration

**Warnings Addressed**
 - Switched use of AlertIOS for updated Alert
 - Looked into warning about RCTNetInfo being deprecated. Unfortunately, trying to update this to the extracted one resulted in a number of difficulties with respect to linking and installing. These are documented in as-yet-unaddressed issues for react-native on github, so despite the warning, this is not yet ready for update.

 
- [x] The README can be updated to include react-native run-ios as a valid command to run the app.

# Review Checklist

- [x] Are tests passing?


wgranger:
I assume this is still a draft and not in need of immediate review?

chelseatroy:
Correct!

mcbouslog (approved):
👍 👍

PR #261, opened 2019-06-12
Quick fix so that Variable Stars shows up in prod. Ultimately we want…

chelseatroy:
… to fix this with pagination.

# Review Checklist

- [x ] Does it work in Android and iOS?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected? <--see README
- [x ] Are tests passing?

### DIFF
--- a/src/actions/projects.js
+++ b/src/actions/projects.js
@@ -6,6 +6,7 @@ import * as ActionConstants from '../constants/actions'
 import { isValidMobileWorkflow } from '../utils/workflow-utils'
 
 const productionParams = {
+    page_size: 50,
     mobile_friendly: true,
     launch_approved: true,
     live: true,


wgranger:
Yep, this fix does bring Variable Stars, and more than 20 `mobile_friendly` projects into the app. Unfortunately, I don't think we can easily use the `cards: true` flag without restructuring several parts of the code.

With the way the app currently works, after fetching a collection of projects, the projects are then sorted into their respective lists given the `isPreview` or `beta_approved` tags ([see here](https://github.com/zooniverse/mobile/blob/master/src/reducers/projectsReducer.js#L34)).

Below is a snapshot of the Variable Star Zoo resource with `cards: true` 
Screen Shot 2019-06-13 at 9 31 22 AM

It could be that `cards: true` will work, but not without rejigging the code in several places. In the meantime, I'll merge this so Variable Stars can be public.

camallen:
If i modify the API cards route to return resource fields like `beta_approved` `launch_approved` etc would that make the changes in the app super easy to make? 

FYI those attributes are super cheap to serialize on the cards route, the main cost is enumerating the relation links.

PR #262, opened 2019-06-18
Variable stars not showing up

chelseatroy:
Variable stars are not showing up because it is on the second page of project results, and we were only fetching the first page.

We now paginate through all the pages of production projects in order to get these. 

In the future, we may also want to paginate through the other types of projects. 

# Review Checklist

- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected? <--see README
- [x] Are tests passing?


src/actions/projects.js
```diff
@@ -39,34 +38,43 @@ const collaboratorParams = {
...
+                }
+
+                projectCalls.push(fetchPaginatedProjects(productionParams));
 
                 // Fetch Beta Projects
-                projectCalls.push(apiClient.type('projects').get(betaParams).then( projects => {
+                projectCalls.push(apiClient.type('projects').get(betaParams).then(projects => {
                     const taggedProjects = tagProjects(projects, false)
                     allProjects = allProjects.concat(taggedProjects)
                 }))
```
wgranger: I think we can refactor lines 66-69 to be `projectCalls.push(fetchPaginatedProjects(betaParams))`

wgranger:
This is about ready to merge. I did encounter a couple errors with this, but running `npm install react-native-vector-icons --save` and `react-native link react-native-vector-icons` fixed my issues, which is weird since the package was already installed, but it might've been something on my end. 

Can you double check if the Euclid project is showing up twice on your end in the Space section?

wgranger:
Looks like Euclid showing up twice is a bug independent of this PR. Merging.

PR #266, opened 2019-07-01
Paginate workflows

chelseatroy:
Describe your changes.

We paginated projects and moved forward under the assumption that workflows were fetched on a per-project basis. As it happens, this is not how the client works: it is fetching them in bulk for all the projects to be show on mobile.

While it is unlikely that any one project has more than 20 workflows, it is not only likely but, in fact, the case that the totality of projects meant to show up on mobile have more than 20 workflows. In order to get more projects into the app, we need to account for this.

# Review Checklist

- [x] Are tests passing?


camallen:
This comment is not blocking but an explanation of how the API provides information to handle paged response data sets. 

The `meta` section of the JSON API response contains paging information. E.g. for the notes from API request for active workflows with project_id = 1588 (NFN), note this workflow didn't modify the `page_size` param so default is 20 and only has 12 worfklows, so `next_href` is null. 

Screenshot 2019-07-02 at 14 04 29

The current implementation here is linked to the default `page_size === 20` and if that ever changes in the API (unlikely) it may change the behaviour of the code here, i.e. stop paging. 

An example of how to page through the data data in ruby is at https://github.com/zooniverse/panoptes-client.rb/blob/3bd483c20a94a70fac438de0fbe0661faa9b34cb/lib/panoptes/endpoints/json_api_endpoint.rb#L11

Finally you can increase the `page_size` param but this only works up to a certain point where it hits a hard limit imposed by the API. We implemented this as some workflows are very large (serialize large json blobs of stats data for wiring selectors (ask Mark). These are costly to serialize from the db to API and also result in MBs of data down the pipes, not ideal for some mobile users ($$$).

Hope this helps, happy to pair / collaborate on how the API provides paging information via the response's `meta` attribute.

wgranger:
Good point Cam made above. I believe an implementation of that approach from PFE is available [here](https://github.com/zooniverse/Panoptes-Front-End/blob/21cf42485929a62938112d5e2d4bca1a6702b00e/app/pages/projects/project-filtering-interface.jsx#L72).

Again, I don't think it is blocking, but it is another example of how we do paging. If you search for `getMeta` in the PFE repo, then you can find other implementations of this.

ios/ZooniverseMobile/Info.plist
```diff
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.4.2</string>
+	<string>2.6.0</string>
```
wgranger: Are we on `2.6.1` now?

src/actions/projects.js
```diff
@@ -119,16 +118,27 @@ const getAvatarsForProjects = projects => {
...
+                    workflow.mobile_verified = workflow.mobile_friendly && isValidMobileWorkflow(workflow)
+
+                    const project = projects.find(project => project.id === workflow.links.project)
+                    if (!project.workflows.find((projectWorkflow) => projectWorkflow.id === workflow.id)) {
+                        project.workflows = R.append(workflow, project.workflows)
+                    }
+                })
+
+                if (workflows.length === 20) {
+                    var currentPage = _page + 1
```
wgranger: I think this should be `newPage`, but this is not at all blocking. However, if you're changing the iOS version string, you might want to change this as well.

wgranger:
I think this looks good. Coding logic appears to work well. Before I approve, is there a particular workflow that is now appearing in the app due to these changes that I can confirm locally?

ios/ZooniverseMobile/Info.plist
chelseatroy: As of the latest, yes. That gets updated by the script, so we have to set it one patch/minor lower than we're at in the code.

chelseatroy: I've been handling this by putting it one lower in the code, then running the script, then updating the code and pushing.

chelseatroy:
@wgranger yes! Radio Meteor Zoo in the space section!

chelseatroy:
> This comment is not blocking but an explanation of how the API provides information to handle paged response data sets.
> 
> The `meta` section of the JSON API response contains paging information. E.g. for the notes from API request for active workflows with project_id = 1588 (NFN), note this workflow didn't modify the `page_size` param so default is 20 and only has 12 worfklows, so `next_href` is null.
> 
> Screenshot 2019-07-02 at 14 04 29
> 
> The current implementation in this PR is linked to the default `page_size === 20` and if that ever changes in the API (unlikely) it may change the behaviour of the code here, i.e. stop paging.
> 
> An example of how to page through the data data in ruby is at https://github.com/zooniverse/panoptes-client.rb/blob/3bd483c20a94a70fac438de0fbe0661faa9b34cb/lib/panoptes/endpoints/json_api_endpoint.rb#L11
> 
> Finally you can increase the `page_size` param but this only works up to a certain point where it hits a hard limit imposed by the API. We implemented this as some workflows are very large (serialize large json blobs of stats data for wiring selectors - ask Mark). These are costly to serialize from the db to API and also result in MBs of data down the pipes, not ideal for some mobile users ($$$).
> 
> Hope this helps, happy to pair / collaborate on how the API provides paging information via the response's `meta` attribute.

Thank you @camallen ! Let's definitely pair at some point. This release needs to go out I think, but it's helpful to see this next_href option, and I'd like to implement it for the next release :)

wgranger:
Ah yeah, Radio Meteor Zoo does fetch in the app but it isn't listed as "mobile friendly."  These changes appear to take care of that and the workflow is now in the expected "made for mobile" section. 

Merging.

PR #279, opened 2019-07-24
Allow Flipbook

### LINKED ISSUE #278: Enable Flipbook into the App
Issue first noticed in a [talk post](https://www.zooniverse.org/talk/17/1067206?comment=1753742&page=1).

Originally, the flipbook was restricted from the mobile app as we couldn't support multi-image subjects. However, we now allow multi-image subjects, and the view for them looks shockingly similar to how the flipbook is displayed on the browser (see below).

Mobile App:
Image from iOS

Browser:
Screen Shot 2019-07-24 at 11 10 53 AM

In order for workflows to appear in the app and still allow the flipbook functionality on the browser, we should allow workflows with the flipbook enabled to appear on the app and show the multi-image subject view. This may be as simple as removing the flipbook check in the workflow utils:
https://github.com/zooniverse/mobile/blob/master/src/utils/workflow-utils.js#L40

However, we'll also have to remember to remove the flipbook restriction on the PFE project builder when an app is being marked "mobile friendly."

wgranger:
Fixes #278  .

Invision Mock-ups: 

Describe your changes.
This PR allows a workflow with a flipbook to be enabled in the app. Read more on the issue above. With the introduction of multi-image subjects, it shouldn't be an issue having a workflow with a flipbook in the app.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/utils/workflow-utils.js
+++ b/src/utils/workflow-utils.js
@@ -8,7 +8,7 @@ export const getTaskFromWorkflow = (workflow) => {
     const key = workflow.first_task
     return workflow.tasks[key]
   }
-  
+
 export const getAnswersFromWorkflow = (workflow) => {
     const task = getTaskFromWorkflow(workflow)
     return task.answers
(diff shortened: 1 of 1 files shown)


wgranger:
See staging link on PR below to enable workflow with flipbook on staging project. https://github.com/zooniverse/Panoptes-Front-End/pull/5476

PR #305, opened 2019-11-27
Upgrade App to RN 0.60

wgranger:
Describe your changes.
This PR updates the app to RN 0.60. The biggest change here is that most pods are now auto linked. The readme will likely need to be rewritten after this merge. The main change being, running from the root `cd ios && rm -rf Podfile.lock && pod install && cd ..` to install the pod dependencies before running the app. The massive massive line removal here is due to the `Pods` folder now being gitignored.

I also want to draw attention to the `signingConfigs` in the `app/build.gradle` on this [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.59.9&to=0.60.0). Those changes are _not_ present in this PR, but we may want to add them should be decide to use the default debug config introduced by RN 0.60. 

Aside from that, in testing this branch please double check Firebase messaging is working, as it was upgraded to version 6 in this PR to take advantage of autolinking. I believe there's info in the wiki on sending a test notification using Firebase. 

Aside from that, if you're able to do an iOS and Android build from this, everything should be good. After merging this, the next step is to upgrade RN further and get some of those dependencies updated.

# Review Checklist

- [x] Are tests passing?


chelseatroy:
Hi!

"The readme will likely need to be rewritten after this merge."

Agreed that the readme needs a lot of changes; let's go ahead and change the README as part of this PR, and not merge until we have that change. For a few reasons:

1. Will make it easier for myself and others to replicate the new build process and confirm that everything is working
1. Prevents us from forgetting to do it, or forgetting something while doing it, and making it really hard to figure out later after it's been a while since we made these changes.

wgranger:
I'm happy to change the README, but I'm curious what the process is like spinning up this app for review.

It might be as easy as adding a change in the readme to `pod install` and `npm run android`, but if there are any other gotchas when reviewing the app, that will also need to be added to the readme. Either way, let me know what the process is like of running this PR and I'm happy to change the readme accordingly.

chelseatroy:
Totally. I've been heads down on theia lately, but I'll see if I can't pick this up over the weekend @wgranger and document my process for you!

PR #306, opened 2019-11-27
Only Mark Museum Mode if User Has Role

### LINKED ISSUE #303: Anonymous users are entering museum mode
Anonymous users are able to enter museum mode. 

**To replicate**: Open the app and visit the mapping historic skies project under the "space" discipline. The app will go into the locked dark mode.

**To Fix**: Currently, the app enters museum mode if _any_ user attached to a project has the `museum` role. Instead, this should check that the user with the `museum` role matches the logged-in user.

https://github.com/zooniverse/mobile/blob/282603301f9260d956e6771fcb02bf3c7c46d02e/src/actions/projects.js#L130

wgranger:
Fixes #303 

Invision Mock-ups: 

Describe your changes.
- Optimize the way we set museum mode. Instead of a call for all `project_roles` on every project in the mobile app, we make a single call for all projects where the current user has the `museum` role, but only if logged in.
- For those projects with the museum role, the `in_museum_mode` flag is triggered if they are in the app.

When reviewing, you can test this out with the Project With All Mobile Workflow Types, which I've promoted to Beta Review and should be available to both logged in and anonymous users.

# Review Checklist

- [x] Are tests passing?

### DIFF
--- a/src/actions/projects.js
+++ b/src/actions/projects.js
@@ -83,7 +83,7 @@ export function fetchProjects() {
                     let projectDetailCalls = []
                     projectDetailCalls.push(getWorkflowsForProjects(allProjects))
                     const avatarCall = getAvatarsForProjects(allProjects)
-                    const museumModeCall = tagMuseumRoleForProjects(allProjects, userProfile)
+                    const museumModeCall = tagMuseumRoleForProjects(allProjects)
 
                     projectDetailCalls = projectDetailCalls.concat(avatarCall)
                     if (userIsLoggedIn) {
@@ -123,7 +123,7 @@ const getAvatarsForProjects = projects => {
     })
 }
 
-const tagMuseumRoleForProjects = (projects) => {
+export const tagMuseumRoleForProjects = projects => {
     return apiClient.type('projects')
       .get({ current_user_roles: 'museum' })
       .then((museumProjects) => {
--- a/src/actions/projects.js
+++ b/src/actions/projects.js
@@ -83,10 +83,12 @@ export function fetchProjects() {
                     let projectDetailCalls = []
                     projectDetailCalls.push(getWorkflowsForProjects(allProjects))
                     const avatarCall = getAvatarsForProjects(allProjects)
-                    const museumModeCall = getMuseumRoleForProjects(allProjects)
+                    const museumModeCall = tagMuseumRoleForProjects(allProjects, userProfile)
 
                     projectDetailCalls = projectDetailCalls.concat(avatarCall)
-                    projectDetailCalls = projectDetailCalls.concat(museumModeCall)
+                    if (userIsLoggedIn) {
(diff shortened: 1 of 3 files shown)


src/actions/projects.js
```diff
@@ -121,16 +123,15 @@ const getAvatarsForProjects = projects => {
...
-            .get({id: project.links.project_roles})
-            .then((roles) => {
-                roles.forEach(role => {
-                    project.in_museum_mode = project.in_museum_mode || role.roles.includes('museum')
-                })
-            })
-    })
+export const tagMuseumRoleForProjects = projects => {
+    return apiClient.type('projects')
+      .get({ current_user_roles: 'museum' })
```
chelseatroy: This looks great! I wish I had known that `current_user_roles` was an option when I was writing this the first time. Is there some documentation or another way that I can figure out things like this in the future?

chelseatroy:
Love the tests! Made one small change to the tests. Also asked a question above, but will merge next week.

src/actions/projects.js
wgranger: It should probably be in the Apiary, but doesn't appear to be there. We're also using `current_user_roles` at the top of this script, but I'm not sure where/if documentation on this lives anywhere.

srallen: Unfortunately it's an undocumented Panoptes feature. It works on any controlled resource, so projects, organizations, and collections for an authenticated requests. The filter should include an array of the user roles you want to filter against. Here's its code if that is at all helpful: https://github.com/zooniverse/Panoptes/blob/4cfe0e6e8bf7b3edd98490416caedb116207753a/app/controllers/concerns/filter_by_current_user_roles.rb

And an example of its use on PFE:
https://github.com/zooniverse/Panoptes-Front-End/blob/21cf42485929a62938112d5e2d4bca1a6702b00e/app/pages/home-for-user/recent-collections.jsx#L48

chelseatroy: Thank you @srallen !

camallen: > Unfortunately it's an undocumented Panoptes feature.

This is unfortunate, looks like we failed to add the docs after this came in, https://github.com/zooniverse/Panoptes/issues/864

FWIW - apiary  got bought by oracle a while back, i'm not sure we've still got access to update it either :( - though feel free to try and update https://github.com/zooniverse/Panoptes/blob/7d8a941fd7bc190e8dc4646da20b11d8b0a251a1/apiary.apib#L1122 to add the filter examples in. If that merges in theory the docs will update on Apiary. 

FYI - The plan is to shift to our own documentation scheme in https://github.com/zooniverse/Panoptes/pull/3050 but we haven't had any traction on porting those docs.

> It works on any controlled resource, so projects, organizations, and collections for an authenticated requests.

And to clarify - it only works on `projects`, `organizations`, and `collections`.

Issue #310, opened 2019-12-03
Crashing Bug: Double Submit


wgranger:
There's a crashing bug in the app when someone presses the "submit" button quickly. The app will turn to a white screen.

**To Replicate:**
Open the Mapping Historic Skies app, which should be in museum mode. Start pressing the "submit" button quickly, and the screen should go all white. The app will need to be rebooted to continue.

chelseatroy:
Fixified :) https://github.com/zooniverse/mobile/pull/311

wgranger:
For context, this is the error we're experiencing. 

![Screen Shot 2019-12-04 at 11 31 40 AM](https://user-images.githubusercontent.com/14099077/70166105-ba4a8d00-1689-11ea-8db1-0c07ecc1ae50.png)

PR #311, opened 2019-12-03
Fix Museum Mode Fast Tap Issue (appears to be specific to drawing workflow)

chelseatroy:
Children can mash the submit button all they want, but it disables after the first press until a new image loads

+ I also checked whether other workflows have this problem. I was not able to reproduce the issue in the other workflows.

Fixes https://github.com/zooniverse/mobile/issues/310.

Invision Mock-ups: 

Describe your changes.

# Review Checklist

- [x] Are tests passing?


wgranger:
LGTM. Nice and succinct fix. I've posted a screenshot of the error message from the simulator in the original issue.

Issue #148, opened 2018-02-06
Sign In screen: Continue without signing in


noahmalmed:
Continue without 'signing in' becomes unhighlighted if text in text fields. 

Should just stay highlighted

beckyrother:
This also needs a Back button `<` in the top left nav bar when you come to this page from `Sign In / Register` in the hamburger menu. Back should take the user back to whatever page they were last looking at.

noahmalmed:
To get started on this, I would check out `SignIn.js`

chelseatroy:
This doesn't happen anymore.

PR #317, opened 2019-12-31
Don't Disable Continue without Signing In Button

wgranger:
Moving towards #148 

Describe your changes.
This is the first step in closing the issue above. It is going to take a bit more work to take care of the second item in this issue (bringing back the < button on the sign in screen). Adding `showBack: true` [here](https://github.com/zooniverse/mobile/blob/master/src/components/SignIn.js#L77) initially takes care of it, but there is a use case where a logged-in user signs out, views the sign in screen, hits the back button, and would see a screen without disciplines.

# Review Checklist

- [x] Are tests passing?


mcbouslog:
Issue #148 closed, closing this PR as well.

PR #327, opened 2020-07-29
Add sentry for error logging

### LINKED ISSUE #322: Use Sentry to Track Bug Reports
While we can get crash counts from the Apple Developer portal we do not get a reason for the crash. Using Sentry (as we use with other Zooniverse apps) would help us get a stack trace to determine why we're experiencing crashes so we can try to resolve them.

chelseatroy:
Fixes #322 

### The Goal: Have a way to remotely receive information about errors and crashes that our volunteers and researchers experience.

Right now, we don't have any info about the cause of issues until and unless we can recreate the issue locally. Some kind of error reporting solution can help us with that, and we're going with Sentry since we already have a Sentry account and already use it to track 6 or 7 other apps for our organization.

+ Here is our brand new Sentry mobile-app project page: https://sentry.io/organizations/zooniverse-27/issues/?project=5371400
+ My thinking was to put the DSN in an environment variable rather than commit it straight to the repo, but the official guidance from Sentry appears to be not to do that: https://forum.sentry.io/t/dsn-private-public/6297

**I have checked the following, but please also check the following:**

- [x] Are tests passing?

**A note on the changes:**

It _looks_ like a lot, but most of these are automated changes that were made by the Sentry dependency setup wizard (I followed [this setup guide](https://docs.sentry.io/platforms/react-native/)). For example, the sentry.properties file and anything in any of the dependencies files was automated. 

The only _manual_ changes in this PR are as follows:

1. `.travis.yml` : Bumped node on the build machine from 8.3 to 8.5 because certain elements of Sentry are only supported at 8.5 and up
2. Changed the post-install script in two ways:
 - `package.json`: Removed "patch-package" because, as of the RN version upgrade, we no longer have to patch react-native-vector-icons
 - Instead we have to mock out 'Icon' for testing, so now _that's_ in the post-install script instead.
3. `containers/app.js`: Add the lines initializing Sentry and pointing it at our DSN for the mobile app.

**What this change should allow you to do:**

1. The app needs to throw an exception when you open it. Mine always does because my profile does not have a profile image, and that throws an exception 😂 . If your profile has a profile image, then you can generate a crash by going into the `containers/app.js` file and adding `Sentry.nativeCrash()` at the top of the `render()` method before it returns. 
2. Run the app on Android or iOS. 
3. Visit our Sentry project page: https://sentry.io/organizations/zooniverse-27/issues/?project=5371400. There, you should see a record of the crash you just caused! 

**My goals for this PR:**

- Honestly, just to get Sentry into the app. We haven't devoted a ton of time to getting specific or fancy with our Sentry usage for mobile, and I think that that would be a separate card, probably to be prioritized after some of the refactors. We just wanted to get _some_ kind of error reporting into the app so that we would have some insights into why crashes happen.
- Explore the use of environment variables on React Native for things like the Sentry DSN. I accomplished this, though it isn't reflected in the code because the Sentry DSN specifically needs to be available before runtime.

### DIFF
--- a/android/sentry.properties
+++ b/android/sentry.properties
@@ -1,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=zooniverse-27
+defaults.project=mobile-app
+auth.token=906b01e443c44171a3fb2a9ef0bddb1d839fe8ce56634d04afe19ad1d5924258
--- a/android/sentry.properties
+++ b/android/sentry.properties
@@ -1,4 +1,5 @@
 defaults.url=https://sentry.io/
 defaults.org=zooniverse-27
 defaults.project=mobile-app
-auth.token=906b01e443c44171a3fb2a9ef0bddb1d839fe8ce56634d04afe19ad1d5924258
+auth.token=See Instructions: https://github.com/zooniverse/mobile/wiki/Enabling-Sentry-for-your-Local-Builds
+
--- a/ios/sentry.properties
+++ b/ios/sentry.properties
@@ -1,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=zooniverse-27
+defaults.project=mobile-app
+auth.token=906b01e443c44171a3fb2a9ef0bddb1d839fe8ce56634d04afe19ad1d5924258
--- a/ios/sentry.properties
+++ b/ios/sentry.properties
@@ -1,4 +1,5 @@
 defaults.url=https://sentry.io/
 defaults.org=zooniverse-27
 defaults.project=mobile-app
-auth.token=906b01e443c44171a3fb2a9ef0bddb1d839fe8ce56634d04afe19ad1d5924258
+auth.token=See Instructions: https://github.com/zooniverse/mobile/wiki/Enabling-Sentry-for-your-Local-Builds
+
--- a/src/actions/projects.js
+++ b/src/actions/projects.js
@@ -42,7 +42,8 @@ export function fetchProjects() {
     return (dispatch) => {
         return new Promise((resolve, reject) => {
             dispatch(addProjectsRequest)
-            getAuthUser().then((userProfile) => {
+            getAuthUser()
+                .then((userProfile) => {
                 const userIsLoggedIn = userProfile !== null
                 let projectCalls = []
                 let allProjects = []
@@ -105,6 +106,8 @@ export function fetchProjects() {
                         dispatch(addProjectsFailure);
                         reject(error)
                     })
+            }).catch(() => {
+                // Stub out avatar rejection because it is optional for users to have avatars
             })
         })
     }
--- a/src/containers/app.js
+++ b/src/containers/app.js
@@ -37,6 +37,12 @@ import SafeAreaContainer from './SafeAreaContainer'
 import { setPageShowing } from '../actions/navBar'
 import NavBar from '../components/NavBar';
 import PageKeys from '../constants/PageKeys'
+import * as Sentry from '@sentry/react-native';
+
+Sentry.init({
+    dsn: 'https://334e2b2ca1c04dc4a7fc356e394e9ea8@o274434.ingest.sentry.io/5371400',
+    enableNative: process.env.NODE_ENV === 'production' ? true : false,
+});
 
 const persistConfig = {
   key: 'root',
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,9 @@ sudo: false
 
 language: node_js
 node_js:
-  - "8.3.0"
+  - "8.5.0"
 
 install:
-  - npm install @sentry/cli
   - npm install
   - npm install react-native-vector-icons --save
   - sed -i -- 's/class Icon extends PureComponent {/class Icon extends PureComponent {constructor(props) { super(props) }/g' node_modules/react-native-vector-icons/lib/create-icon-set.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ node_js:
   - "8.3.0"
 
 install:
+  - npm install @sentry/cli
   - npm install
   - npm install react-native-vector-icons --save
   - sed -i -- 's/class Icon extends PureComponent {/class Icon extends PureComponent {constructor(props) { super(props) }/g' node_modules/react-native-vector-icons/lib/create-icon-set.js
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -80,6 +80,7 @@ project.ext.react = [
 ]
 
 apply from: "../../node_modules/react-native/react.gradle"
+apply from: "../../node_modules/@sentry/react-native/sentry.gradle"
 apply from: "../../node_modules/react-native-vector-icons/fonts.gradle"
 
 /**
--- a/ios/ZooniverseMobile.xcodeproj/project.pbxproj
+++ b/ios/ZooniverseMobile.xcodeproj/project.pbxproj
@@ -5,7 +5,6 @@
 	};
 	objectVersion = 46;
 	objects = {
-
 /* Begin PBXBuildFile section */
 		00E356F31AD99517003FC87E /* ZooniverseMobileTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 00E356F21AD99517003FC87E /* ZooniverseMobileTests.m */; };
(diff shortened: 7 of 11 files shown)


chelseatroy:
TODO: 

[] Fix Tests
[] Fix Build
[] Switch between development, staging, and production from launch

chelseatroy:
To get the env var I tried using `react-native-dotenv` like so:

: https://levelup.gitconnected.com/how-to-gracefully-use-environment-variables-in-a-react-native-app-7f1600446116

Turns out that injects the URL at _runtime_ and Sentry wants to see it at _build_ time, so the app won't build by this method. 

Next, I'm going to go the old-fashioned route: export it from a file that we don't push to Github and only use for builds, then add instructions to the README for how to get that info. 

The finished product would require reverse-engineering the app to get the URL—which someone could absolutely do, but:

1. An env var actually would not have fixed that
2. It's still much more onerous than trawling public github repos for keys.

chelseatroy:
OK, my final idea didn't work, and Cam and Jim mentioned that we do already have Sentry keys in other public repos. I don't love it, but I think for now I'm gonna go with what the Sentry maintainers say and leave it in.

android/sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=zooniverse-27
+defaults.project=mobile-app
+auth.token=906b01e443c44171a3fb2a9ef0bddb1d839fe8ce56634d04afe19ad1d5924258
```
wgranger, line 4: Should this be hidden? I saw the note about the DSN being in the repo, but having checking this into version control seems strange to me, unless the docs say otherwise.

ios/sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=zooniverse-27
+defaults.project=mobile-app
+auth.token=906b01e443c44171a3fb2a9ef0bddb1d839fe8ce56634d04afe19ad1d5924258
```
wgranger, line 4: Same note here as in the gradle

src/containers/app.js
```diff
@@ -37,6 +37,12 @@ import SafeAreaContainer from './SafeAreaContainer'
 import { setPageShowing } from '../actions/navBar'
 import NavBar from '../components/NavBar';
 import PageKeys from '../constants/PageKeys'
+import * as Sentry from '@sentry/react-native';
+
+Sentry.init({
+    dsn: 'https://334e2b2ca1c04dc4a7fc356e394e9ea8@o274434.ingest.sentry.io/5371400',
+    enableNative: process.env.NODE_ENV === 'production' ? true : false,
+});
```
wgranger, line 45: This definitely works. It's pretty cool Sentry will start reporting with a quick `.init()` at the root. 

However, this is currently reporting non-crashing bugs and warnings. It might be worthwhile to add an option here to only report crashing bugs. More info in the PR review.

chelseatroy: So, this one is interesting. I researched several things, and here's what I found:

1. Sentry has a couple of options that you can include in the `init` to manage what is sent, but none of them is specifically a log level for messages from the app itself. It has:

`logLevel` - this is not actually for the level of logs _from our app_. It is for the level of logs that Sentry should show us if Sentry _itself_ fails to send our app's errors through to its console.

`ignoreErrors` - this allows us to ignore _specific_ errors. That said, it's for errors—not warnings. It also requires us to know, verbatim, what we wish to exclude.

`error_types` - Only exists according to the documentation. I did not find it upon a thorough investigation of the ReactNativeOptions interface and all the interfaces it inherits from (`BrowserOptions`, `Options`). Supposedly, inherits a bytemask system from PHP that allows you to specify exactly which kinds of errors you want (defaulting to `E_ALL`). Like I said, does not exist in the implementation. Only mentioning in case you decide to do your own check of the documentation and you find this.

2. There's a manual approach. We can initialize Sentry with a pre-post block called `beforeSend` as [demonstrated here](https://docs.sentry.io/error-reporting/configuration/filtering/). However:

- It doesn't actually _stop_ Sentry from sending the event. It just allows us to modify it—say, by adding a tag, which theoretically means that the event would then only appear once _in the UI that we see_. I don't know for sure that it affects how each event counts toward our monthly limit.

3. Finally...it actually appears that Sentry _doesn't_ log warnings. It only logs errors. The "no avatar" situation is, it turns out, a full-blown error. It says here `type:error`:

error

Furthermore, when I run the app, I see all of the following warnings, and none of those appear in Sentry:

warnings

So I think the solution here might be to handle this particular error. Do we have any data on what proportion of our users do not have avatar images?

wgranger: Nice! Looks like there's a lot of good stuff covered here to help us figure out a good solution. I'm not sure what proportion of our users have avatar images, but I think we should _assume_ many of them do not. Handling this particular error is probably the best bet here.

I'm thinking there are several other errors hidden throughout the app similar to the avatar error that would log to sentry. Ex: I found another one while reviewing this PR related to signing out. The best solution here may be to add Sentry to the app, and then play around with the app in staging to audit that app and find out what errors are submitted to Sentry. We can then handle those errors in a separate PR as it's out of scope of this PR.

Otherwise, I can see a situation where we breeze through out Sentry allotment quickly without gracefully handling errors.

chelseatroy:
Further investigation apropos of the "no avatar" error.

Turns out, it's happening in the PanoptesClient. Here is where Sentry states the error is happening:



in a `handleError` method. Here's where that method lives:



This gave me the hypothesis that a call to PanoptesClient is throwing somewhere in our code and we're not catching it. So I started looking. 

Turns out...that's exactly what was happening. In fact, we already catch exceptions in the mobile app in several places (see `getAvatarsForProjects` in actions/projects.js` for an example). But we missed one: fetching projects for a user. I added an exception catch on that promise stream, and that solved this particular error.

android/sentry.properties
```diff
@@ -1,4 +1,5 @@
 defaults.url=https://sentry.io/
 defaults.org=zooniverse-27
 defaults.project=mobile-app
-auth.token=906b01e443c44171a3fb2a9ef0bddb1d839fe8ce56634d04afe19ad1d5924258
+auth.token=See Instructions: https://github.com/zooniverse/mobile/wiki/Enabling-Sentry-for-your-Local-Builds
```
wgranger, line 4: Can we add this to the README? I think this is an important step when getting the app up and running with a new dev

chelseatroy: Totally! Lemme do that right now

android/sentry.properties
chelseatroy: YIKES. Yes, that should be hidden. Let me fix that. Thank you for catching that!

src/actions/projects.js
```diff
@@ -105,6 +106,8 @@ export function fetchProjects() {
                         dispatch(addProjectsFailure);
                         reject(error)
                     })
+            }).catch(() => {
+                // Stub out avatar rejection because it is optional for users to have avatars
```
wgranger, line 110: I think this is alright. Normally I would at least add a `console.warn` or something similar here, but that's not as helpful for the mobile app.

ios/sentry.properties
chelseatroy: Same yikes here as above, lol

src/actions/projects.js
chelseatroy: Yeah, in this case I think we're okay because a user having an avatar really _should_ be optional. Having no avatar doesn't affect their ability to engage with the core functions of the app (working on projects), and from a data ethics perspective any user data that we don't absolutely need should be optional.

If it _were_ something that affected some app functionality in some way, that's where I'd throw up a warning.

src/containers/app.js
chelseatroy: I agree—I went ahead and fixed that error (see below comment).

I also did a little manual testing: I went into the app and clicked all around. I looked at tutorials. I looked at field guides. I performed classifications. Then I logged out. I didn't see any more errors.

That said, I _do_ see the error you're mentioning in Sentry from the original review: "Error: Failed to sign out; not signed in." How, uh...where in the UI did you find a sign out button while not signed in? We probably shouldn't have that 😂

wgranger: Cool! Then if that's the only exception that needs to be handled I suppose it can be dealt with in this PR

chelseatroy: Rock on. I tried to reproduce signing out while signed out, but haven't so far. So I'll merge this in, and if we figure out how to repro that, I'll fix it in a new PR!

PR #330, opened 2020-08-27
Inspect warnings

chelseatroy:

Everything in the app should work the same as it did before. These were changes related to getting ahead of deprecations in react native.

# Review Checklist

- [x] Are tests passing?

# My goals for this PR:

1. We have several warnings in our app for things that are going to be deprecated. This PR fixes some of those!


mcbouslog (approved):
Looks good, tested on iOS - iPhone 11, Android - Pixel 2, Pixel 3.

Unsure how to test notification, but code looks good from review of related react docs.

PR #336, opened 2020-09-15
Remove resizing logic in DrawingClassifier.

### LINKED ISSUE #335: Subject image not loading in drawing classifier
In order to fix an issue, we need to be able to reproduce it. To give us the best chance of reproducing your issue, please follow these steps:

1. The version of Android or iOS that you are testing on.
2. The device that you are testing on (for example, iPhone 7 or Samsung Galaxy s9)
- Android Studio Simulator: Pixel 3 API 29
- Android Studio Simulator: Pixel 2 API 29
- my personal Pixel 3, Android v10

3. What you did. Example: “I navigate to Floating forests, click the “field guide” button, then click on one of the options.”
- workflow with a drawing task, the subject image doesn't load
- other workflow subject images load ok

4. What you expected to happen. Example: “I expected the option to open with a detail screen.”
- subject image loads

5. The actual behavior.  Example: “The app crashed.”
- subject image loading sticks on spinner

6. Please include, if relevant, what particular projects or workflows it is happening on.
- https://master.pfe-preview.zooniverse.org/lab/1608/workflows/3437

7. Please include whether you were logged in or logged out, acting as a guest user. If the issue has to do with a particular project or workflow, please try logging in and out and see if the issue persists in both cases.
- occurs logged in or out

NativeImageError

I think the issue is related to this line - https://github.com/zooniverse/mobile/blob/master/src/nativeModules/NativeImage.js#L18
Specifically, I don't see ImageSizer defined in the https://reactnative.dev/docs/native-modules-setup.

Or related, there might be an issue with the cached image url created and passed to NativeImage in the following lines - https://github.com/zooniverse/mobile/blob/master/src/components/Markings/DrawingClassifier.js#L112-L122

I don't think this issue is related to #320 , but is preventing testing on Android.

chelseatroy:
+ It was failing silently on Android (causing images to never load).
+ Tried 3 alternate libraries to fix: none worked.
+ Eventually removed the resizing logic and realized that the app looks exactly the same with or without the resizing logic, which apparently is not used during the actual display.

Fixes #335.

# The Problem:

Currently, images do not load on either the Android emulator or a device (see issue). The problem is related to a resizing API working only on iOS. 

1. Boot up Android emulator OR iOS Simulator.
2. Visit Arts > Plane Staging Project, go through tutorial, and see an image load! OR
Visit Beta > Project with All Workflows, go to the drawing workflow, go through the tutorial, and see ghost images load!

Invision Mock-ups: 

# Review Checklist

- [ ] Are tests passing?

# My goals for this PR:

1. Honestly, get the thing working again. Image loading is **core functionality**, so it is high priority to keep this working on all devices. We also want this working to be able to test new PRs, such as Lance's PR for the deletion buttons.
2. Keep the change small if possible. In this change, we chiefly _remove_ code that evidently was not serving us.


mcbouslog (approved):
This LGTM, great job on a frustrating bug! Tested on iPhone 10, iPhone 11, Pixel 2, and Pixel 3.

Optional items:
- I think we can delete the NativeImage file?

For my own understanding, why don't `height` and `width` need `const` or `let`? If I add `const` it breaks the Android app.

chelseatroy:
> This LGTM, great job on a frustrating bug! Tested on iPhone 10, iPhone 11, Pixel 2, and Pixel 3.
> 
> Optional items:
> 
> * I think we can delete the NativeImage file?

Done! Removed.

> 
> For my own understanding, why don't `height` and `width` need `const` or `let`? If I add `const` it breaks the Android app.

My suspicion is that it has something to do with ES2015+ not compiling in a backward-compatible way to our lowest target APK for Android.

Issue #290, opened 2019-10-25
Allow video subjects in the classifier


wgranger:
As a researcher,
When I have research subjects captured on video,
I should be able to put my project on the mobile app,
So that mobile-based volunteers can help me do my research.

**Here is a screenshot/diagram of what this should look like:** (please put the screenshot here rather than a link to an inVision, because open source contributors do not have access to our private inVision account)

**Here is a project we can test this on:** @lcjohnso please let me know if we have such a project in prod, and/or let's make one in the "mobile app all workflows" sample project on staging!

**Additional context**

Currently, all subjects are rendered through the react native `Image` component. In order to open more projects to the app, we should find a way to render video subjects.

We currently have **X number of projects** in production that could take advantage of this feature.

PR #339, opened 2020-09-22
Enable gifs on Android

chelseatroy:
+ Gifs are not automatically supported on Android in react-native: you have to include a library specifically to support them.

Fixes # 332.

1. Run the Android Emulator (while it is pointing to staging)
2. Go to "Beta Projects"
3. Find the project called "Project with all mobile workflows"
4. Click the "multi answer question" workflow and go through the tutorial. I have the subject set for this workflow set to include gifs!
5. Go through the tutorial, get to the images and ensure that they move!

**This PR is not working if** the images _show up_ but don't _move_.

This was working already, but for peace of mind, please feel free to also try this exact same series of steps on an iOS Simulator to ensure the images move for everybody :)

# Review Checklist

- [x] Are tests passing?

# My goals for this PR:

There's no code here—only a line added to the Android build.gradle file and subsequent dependency file updates—so there isn't much to have goals about. We _did_ want this working because it increases the number of projects that can use the mobile app with the existing workflows until I get around to issue #290 (grazie @wgranger) and add video support.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -175,6 +175,8 @@ dependencies {
     implementation project(':@react-native-community_blur')
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.facebook.react:react-native:+"  // From node_modules
+    implementation 'com.facebook.fresco:animated-gif:2.0.0' //gif support
+
 
     // JSC from node_modules
     if (useIntlJsc) {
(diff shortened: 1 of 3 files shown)


mcbouslog (approved):
LGTM! Tested on Pixel 2, Pixel 3, and confirmed everything still working as expected on iPhone 11 though as noted this PR doesn't affect iOS. Confirmed a few other workflows (swipe, multi-question) still work as expected too.

PR #342, opened 2020-09-28
Sort projects so active ones appear first

### LINKED ISSUE #337: Improve Project Display: Selection + Ordering
The current procedure that volunteers must follow to select projects and workflows to work on in the mobile app is difficult and time consuming: many projects and workflows shown are "out of data", forcing volunteers to search through long lists of projects to find one where their effort is wanted.

**Current Behavior:** The mobile app displays all active, mobile-enabled project + workflow combinations.  Ordering of projects within category-specific view appears to be alphabetical based on project name.  Also, the app's "show all workflows" setting that adds a "In-Browser Experience" section listing non-mobile projects is True by default.

**Proposed Behavior - Major Change Version:** Do not show projects for which all its workflows are out of data.  Equivalent to how "paused" projects (where all active workflows are complete / out of data) are moved to a separate page on the PFE projects page.  Note that the implementation to achieve this goal would necessarily be more complex than filtering projects based on Active vs. Paused status, as there are many cases where a project is Active due to non-mobile workflows, but should be Paused from a mobile-only perspective.  But it appears that there is already logic to decide whether or not to display an "out of data" flag on the project, so perhaps it would be possible to choose not to displays these projects at all.

**Proposed Behavior - Minor Change Version:** Sort projects in each category-specific view so projects with data to be classified (are *not* out of data) appear at the top of the list for best visibility.

**Proposed Behavior - additional change regarding in-browser options:** Set default for "Show all workflows" setting to False in order to restrict the mobile app for use with mobile-friendly workflows only.  Pro: focuses app usage to mobile-specific use case.  Con: new setting would reduce the number of available projects shown in the app.

I'm interested to hear feedback and concerns on this suggestion.
cc: @beckyrother @trouille @chelseatroy 

Related & supersedes: https://github.com/zooniverse/mobile/issues/178

chelseatroy:
Fixes #337  .

1. Open the app pointed at production panoptes (you can do this by setting line 60 in `config.js` in the `panoptes-client` in `npm_modules` to be `var env = "production";` instead of `var env = envFromBrowser || envFromShell || DEFAULT_ENV;`).
2. Go to the "Physics" tab. 
3. The Galaxy Zoo mobile project should appear _before_ all the projects that are out of data. 

Invision Mock-ups: 

# Review Checklist

- [x] Are tests passing?

# My goals for this PR:

1. Once again, this was a "get it working" change rather than a large refactor. There is very little code. I use an anonymous function for the sorting because I couldn't find evidence that we would need to do this same sort anywhere else.

2. Start adding test coverage for these project display options. I ran out the timebox on determining a good way to test them within components, so I began extracting them from components into a `projectDisplay` collection inside of a directory called `displayOptions`. The plan would be to start to extract modifications we run on project lists to there, so we can test them there and import them where we need them!


mcbouslog (approved):
LGTM! Tested on iPhone 11, Pixel 2, Pixel 3.

PR #345, opened 2020-10-10
Fixes rectangle deletion on Android. Also changes the way delete works.

### LINKED ISSUE #341: Can't delete drawn rectangles on Android devices
In order to fix an issue, we need to be able to reproduce it. To give us the best chance of reproducing your issue, please follow these steps:

1. Please check that you currently have the latest mobile app version (2.6.9 as of Dec 3, 2019) installed. If not, please update.
- this issue is on current `master` branch

In your issue, please include:

1. The version of Android or iOS that you are testing on. - API 29
2. The device that you are testing on (for example, iPhone 7 or Samsung Galaxy s9) - Pixel 3 and Pixel 2
3. What you did. Example: “I navigate to Floating forests, click the “field guide” button, then click on one of the options.”
- Navigate to drawing workflow, draw rectangles, click to delete mode, click "X" to delete rectangle
4. What you expected to happen. Example: “I expected the option to open with a detail screen.”
- I expected the rectangles to be deleted
5. The actual behavior.  Example: “The app crashed.”
- nothing happened, no crash, could submit/close and continue using app
6. Please include, if relevant, what particular projects or workflows it is happening on.
- https://master.pfe-preview.zooniverse.org/lab/1608/workflows/3437, or any drawing workflow
7. Please include whether you were logged in or logged out, acting as a guest user. If the issue has to do with a particular project or workflow, please try logging in and out and see if the issue persists in both cases.
- both

chelseatroy:
+ No more transparent button with an x on it that has to move around if a rectangle's edge is near the image's edge.
+ Now, in delete mode, all rectangles go from green to red. Click anywhere on a rectangle to delete it.

Fixes #341.

So, originally, the way that drawing flows worked was that they had insert mode and delete mode. In insert mode, you could draw and move rectangles. In delete mode, a transparent, round button with a green border and a green x appeared in the upper right hand corner.

Well, if a rectangle was up against the right side or top of the image, you couldn't reach the delete button. 

Screen Shot 2020-10-12 at 12 41 09 PM

So an open source contributor submitted a change that would move the button to any of the four corners, depending on where the rectangle was.

Screen Shot 2020-10-12 at 12 41 31 PM

This solved the problem of making delete touch targets available for all rectangles. But it introduced a new problem: the button was now harder to find. Because it is transparent with a green x, it blends into images that have green in them, or that have a lot of detail in them. Plus, now you could not narrow down where it would be based on the rectangle position, because it could be in any of the four corners.

There was another problem: Android and iOS calculate positions slightly differently, and the change was calibrated for iOS screen dimension units. So on Android it didn't work, since Android thought the button was in a different place than it was.

We could have just added position conversions for all four corners for Android, doubling our number of special cases from 4 to 8. Such code would be a pain to maintain, and the feature already wasn't user-friendly.

So I changed it. 

Now, there is no more "remove" button on the rectangles. In insert mode, it's a green rectangle that you can draw and resize, like before.

Screen Shot 2020-10-12 at 12 43 16 PM

In delete mode, all the rectangles turn red, and clicking on one removes it. 

Screen Shot 2020-10-12 at 12 43 24 PM

This way, since at least one corner of the rectangle is _always_ accessible, it is always possible to delete.

Invision Mock-ups: 

# Review Checklist

- [x] Are tests passing?

# My goals for this PR:

1. Simplify the code. We had added edge case after edge case for a while, and it took me several days to identify the problem. I would not want another maintainer to have to do this.
2. Make a friendlier UI. As the developer of the app, I often struggled to find the "remove" button. I cannot imagine that volunteers would have an easier time finding it than I did.
3. Fix the Android situation. Rectangles need to be deletable on the app for _both_ frameworks.


beckyrother:
I really like this solution! Thinking about a11y / colorblindness though, just switching between red and green may not be super obvious for everyone. How do you feel about adding a diagonal line pattern inside the boxes as well, something like: https://projects.invisionapp.com/d/main#/console/13393648/434396833/preview

chelseatroy:
> I really like this solution! Thinking about a11y / colorblindness though, just switching between red and green may not be super obvious for everyone. How do you feel about adding a diagonal line pattern inside the boxes as well, something like: https://projects.invisionapp.com/d/main#/console/13393648/434396833/preview

Oooh, you're right! I forgot to check that. I appear to not have access to that link (no idea why). Could you please post a screenshot of it here?

beckyrother:
Here's the image:
image

Is there a different email that you use for InVision? You've got edit permissions with `chelsea@zooniverse` but I can add another if that's easier

chelseatroy:
@beckyrother here's one thing I can _definitely_ do:

Screen Shot 2020-10-12 at 1 14 34 PM

Let me see what I can find about filling shapes with a pattern!

chelseatroy:
@beckyrother WAIIIIIT I got patterns going. I need to figure out the math for this specific pattern but I got it!

chelseatroy:
For reviewers' reference: I timeboxed the pattern effort, hit the timebox, and fell back on the semi-opaque fill solution.

mcbouslog (approved):
This LGTM! 

I noted an unrelated issue (also on `master`) I'll open a GitHub issue for.

mcbouslog (changes requested):
- [x] for iOS iPhone11 I got an error on drawing workflow load - `naturalHeight` undefined per `MarkableImage.js`. Adding the following default props to `MarkableImage.js` appears to fix:
```
MarkableImage.defaultProps = {
    subjectDimensions: {
        naturalWidth: 200,
        naturalHeight: 240
    },
}
```

- [x] on Android Pixel 3 there was still a little cutoff, but changing the height to `260`(in `DrawingClassifier.js` and `MarkableImage.js`)  seems to help - EDIT - just read commit messages, not sure about height, `240` is definitely better, not sure if downsides to increasing to `260` or `280`, so I'm going to mark this as ok, but def something to revisit (as already noted)

mcbouslog (approved):
Looks/works great on iPhone11, Pixel 2, and Pixel 3!

PR #346, opened 2020-10-17
Fix android dependencies

chelseatroy:
Fixes # .

Build the application! We had some dependency updates to make to get a successful build for release. 

Invision Mock-ups: 

# Review Checklist

- [ ] Are tests passing?


chelseatroy:
We cut a release off this branch and then a bunch of folks pulled it from testflight and tested it before we went to production—so it doesn't have a PR review but it was heavily reviewed!

PR #363, opened 2021-03-27
Add task type checks to isValidQuestionWorkflow and isValidMobileWorkflow tests

### LINKED ISSUE #360: TypeError: undefined is not an object (evaluating 'n.question.length')
I think in [workflow-utils](https://github.com/zooniverse/mobile/blob/master/src/utils/workflow-utils.js#L24) if the check for a valid mobile drawing workflow/task returns false, it then tries to check the workflow/task as a valid mobile question workflow/task, but while checking it errors on task.question (as drawing tasks have task.instruction, as properly checked in the drawing validation function). This error prevents subsequent workflows from properly loading.

I think this issue was causing beta projects (projects, and related workflows who are requested last) to show as Out of Data, as the projects were properly requested and sorted, but the workflows weren't properly loaded in the app.

I edited the workflow/task (https://www.zooniverse.org/lab/9747/workflows/12852) that was causing the issue, but a different drawing workflow/task could cause the issue again without a change to the code. I'll try to draft a PR asap to address code change needed.

Sentry Issue: [MOBILE-APP-3F](https://sentry.io/organizations/zooniverse-27/issues/2183869495/?referrer=github_integration)

```
TypeError: undefined is not an object (evaluating 'n.question.length')
  at isValidQuestionWorkflow (src/utils/workflow-utils.js:37:40)
  at isValidMobileWorkflow (src/utils/workflow-utils.js:24:46)
  at None (src/actions/projects.js:148:76)
  at forEach ([native code])
  at fn (src/actions/projects.js:147:27)
...
(12 additional frame(s) were not displayed)
```

mcbouslog:
Fixes #360.

There's no immediate/live/active issue related to this, but opening for consideration to include in the upcoming release.

Background: invalid drawing task workflows were returning false when validated with the drawing task workflow validation, but then passed to the question task workflow validation which would throw an unhandled error, preventing current and subsequent workflow loading.

- no immediate change to user experience, though #360 was previously preventing workflows from loading, the workflow that caused the issue was manually fixed
- load all valid mobile workflows if a new invalid task workflow is created

# Review Checklist

- [ ] Are tests passing?

# My goals for this PR:
- add task type checks in the question task workflow validation, so an invalid drawing or any invalid task type returns false from the question task workflow validation
- adds tests for valid and invalid question (single and multiple) and drawing task workflows

### DIFF
--- a/src/utils/__tests__/utils-test.js
+++ b/src/utils/__tests__/utils-test.js
@@ -69,6 +69,26 @@ it('gets a subject location ', () => {
   expect(getSubjectLocation(subject)).toEqual(returnSubject)
 })
 
+it('passes a good drawing workflow', () => {
+  let testWorkflow = Object.assign({}, mockWorkflow)
+  testWorkflow.first_task = 'T0'
+  testWorkflow.tasks['T0'] = {
+    help: 'Good luck!',
+    instruction: 'Draw on the cat!',
+    tools: [
+      {
+        color: '#00ff00',
+        details: [],
+        label: 'Rectangle Tool',
+        type: 'rectangle'
+      }
+    ],
+    type: 'drawing',
+  }
+  
+  expect(isValidMobileWorkflow(testWorkflow)).toBeTruthy()
+})
+
 it('passes a good single question workflow', () => {
   let testWorkflow = Object.assign({}, mockWorkflow)
   testWorkflow.first_task = 'T0'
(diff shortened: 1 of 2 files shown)


chelseatroy:
Hi @mcbouslog—thank you for taking care of this!

Should this get QA'd for the release I'm doing this weekend?

mcbouslog:
@chelseatroy - it does not have to be included in the release, as even if the issue occurred again we could manually edit the problem workflow causing the issue. As this (Android) release is for a very particular purpose I lean towards not including this. 

That being said, I'm not sure how much including this PR for release would complicate things? I'm assuming it's more than minimally, therefore better to wait and include in next full release, but if that assumption is wrong then we could include it.

Sorry, long story short - I think best to wait to include this in the next full release, but open if think otherwise.

chelseatroy (approved):
Looks good!

PR #383, opened 2021-10-30
Disable build system deprecation error in xcode

mcbouslog:
When running the app using XCode I hit an error related to the build system will be deprecated. 

This error can be resolved by changing the following setting within Xcode (bottom left red arrow):
Screen Shot 2021-10-30 at 4 20 44 PM

I think updating `ios/ZooniverseMobile.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings` accordingly will automatically enable the setting noted for this project.

This change is not necessary, but might make it easier (one less thing to worry about to get up and running) for other devs to get started.

# Review Checklist

- [ ] Are tests passing?

# My goals for this PR:


chelseatroy (approved):

chelseatroy:
@mcbouslog excellent catch! Thank you for this; this change is a great addition for helping folks get up and running developing our app with minimal hassle :)

PR #398, opened 2022-05-05
Upgrade `react-native-google-analytics-bridge` to v7.1.0

mcbouslog:
I was able to get the `video-subjects` branch running for Android after editing the `node_modules/react-native-google-analytics-bridge/android/build.gradle` code to use `implementation` instead of `compile`.

The (deprecated) dependency fixes the issue in v7.0.0 ([link to repo tag](https://github.com/idehub/react-native-google-analytics-bridge/releases/tag/v7.0.0)).

However, after upgrading, if I attempt to refactor the related code to [use as suggested](https://github.com/idehub/react-native-google-analytics-bridge/releases/tag/v7.0.0), tests fail with `_reactNativeGoogleAnalyticsBridge.GoogleAnalyticsTracker is not a constructor`.

I can dig into this more, but first want to check if this helps @chelseatroy run the app for Android locally?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21764,9 +21764,9 @@
       }
     },
     "react-native-google-analytics-bridge": {
-      "version": "5.9.0",
-      "resolved": "https://registry.npmjs.org/react-native-google-analytics-bridge/-/react-native-google-analytics-bridge-5.9.0.tgz",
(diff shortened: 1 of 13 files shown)


mcbouslog:
I'll follow-up regarding the deployment in Slack.

Closing in favor of #402 .

Issue #281, opened 2019-08-21
Update GA to use Firebase SDK for Continued Tracking


wgranger:
We use GA in the mobile app to track page views. However, I'm not sure who is using this data currently. Nonetheless, GA is sunsetting their older GA SDK in favor of Firebase. **This will happen on October 31 when they stop tracking such properties.** 

To get started:
- [Create a brand new App property in Google Analytics](https://www.google.com/appserve/mkt/p/AFnwnKWbQeeF843rPjpJHY1z1hb1QRr390mjp9jWJrHHfGALytaNsBGhRirYhZxIVvpGHMqxBemi_OxiuFBqPt42kFuurtu0XaTrzK1ubvdCHVJJlbXhV-o0pYg8grnjf2op)
or
- [Upgrade your existing Firebase Project to the full Google Analytics experience](https://www.google.com/appserve/mkt/p/AFnwnKXJnBe1QDNvgQ0sRMbF3_R2_JjkPvKaWic_xukzalBf0rDiG0gPZv5Wm91HNFxnVT7gcAbi_3hkHtza7BqFBEalMCy37hCRk_g9ayV7TDmQqIUgXu68)

lcjohnso:
Hi @wgranger (and others): what is the current status of mobile analytics?  Is there currently no data collection due to sunsetting?

lcjohnso:
Related: https://github.com/zooniverse/mobile/issues/172

lcjohnso:
Closing, but see #413

Issue #172, opened 2018-04-12
Add google analytics for classifications


noahmalmed:
Fix pending NSF grant

noahmalmed:
Also as a part of this, we need to migrate to google analytics new platform. Per this email from google: 

> Dear Google Analytics customer,
> 
> You are receiving this email because you are in the first wave of customers who have been identified as users of the older style of Google Analytics for mobile apps reporting and the Google Analytics Services SDK. Specifically, we are notifying you regarding your Google Analytics property(ies) (, ID: UA-1224199-62: "Zooniverse Mobile").
> 
> We want to let you know that in October 2019 we will begin to sunset our Google Analytics for mobile apps reporting and the Google Analytics Services SDK. We are investing our resources in the latest style of app reporting in Google Analytics that works in conjunction with Firebase – Google's integrated app developer platform. As such, the following will take place:
> 
> In 2019, we will begin to decommission properties that receive data exclusively from the Google Analytics Services SDK.
> Data collection and processing for such properties will stop on October 31, 2019.
> Reporting access through our UI and API access will remain available for these properties' historical data until January 31, 2020.
> After our service is fully turned down, these properties will no longer be accessible via our Google Analytics UI or API, and their data will be removed from Google Analytics servers. You will receive further notification when this time nears.
> At this time, no Analytics 360 properties are impacted by these changes.
> 
> We want to give you plenty of time to make the transition. The good news is that the latest solution using the Firebase SDK is even more intuitive and includes free and unlimited event reporting to meet the needs of app-centric businesses. We've invested heavily to make this solution best-in-class, with new features and capabilities rolling-out continually. Additionally, our offering is closely integrated with other Google products and features to help grow your app business like Crashlytics, AdMob and Remote Config.
> 
> Getting started with our latest app reporting features is simple and straightforward. Here's how. For additional information on our new Google Analytics app reporting, visit the Help Center.
>

lcjohnso:
Closing -- now see #413

PR #402, opened 2022-06-29
Remove Google Analytics and `react-native-google-analytics-bridge`

mcbouslog:
Fixes issues running app locally with Android Studio.

- run the app locally with Android Studio (Android Studio Chipmunk | 2021.2.1 Patch 1, node 10)

# Review Checklist

- [ ] Are tests passing?


mcbouslog:
These changes allow the app to run locally with Android Studio. Google Analytics isn't used, so this PR removes it and the related deprecated dependency that implemented it. See #281 and #172 for additional information regarding next steps for analytics.

Noting per #398 you don't have to run this locally but I think GitHub requires an approving review to be able to merge.

PR #403, opened 2022-07-05
Build and archive changes for iOS v2.9.1

mcbouslog:
These are changes per the iOS v2.9.1 deploy.
I think maybe the changes for `project.pbxproj` and `ZooniverseMobile.xcscheme` should be added to `master`, but maybe not the (version) change to `Info.plist` (which should be kept as `MARKETING_VERSION`)?

# Review Checklist

- [ ] Are tests passing?

# My goals for this PR:


mcbouslog:
Actually I don't think any of these changes are substantive, can likely close this PR (without merge).

PR #407, opened 2022-09-08
Set app in light mode

mcbouslog:
Fixes issue with typed text not showing (i.e. in username and password inputs).

# Review Checklist

- [ ] Are tests passing?


lcjohnso (approved):
LGTM!

PR #410, opened 2022-09-09
Refactor SwipeCard to hide SubjectOptionsBar for video subjects on iOS

mcbouslog:
Fixes issue on iOS with video subjects when click Zooniverse expand button (not iOS fullscreen button), then click iOS minimize button (not Zooniverse "X" button), then on returning to normal classifier the video controls are gone and classifier buttons don't work.

# Review Checklist

- [ ] Are tests passing?


lcjohnso (approved):
LGTM!

PR #417, opened 2022-09-12
Build changes for iOS v2.9.2 release

mcbouslog:


ios/ZooniverseMobile/Info.plist
```diff
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>$(MARKETING_VERSION)</string>
+	<string>2.9.2</string>
```
lcjohnso, line 20: Was it necessary to eliminate this variable, or was this your preference?  Personally, I like setting the value directly in Info.plist, but then can we also remove the two commands to set the variable in `project.pbxproj`?

Also: we should change the iOS build instructions [GDoc link](https://docs.google.com/document/d/1kMbryj3tvJhnXkdIgHcnmOJQqwh_VMLspuCFYJfn_0s/edit) to match.

lcjohnso (approved):
Good to go, but one comment to future-proof iOS deploys and avoid confusion.

ios/ZooniverseMobile/Info.plist
mcbouslog: I was able to delete the references to `MARKETING_VERSION` from project.pbxproj and ran the app locally and archived/built the app for release without any noticeable issue (though I did not upload archive to App Store).

mcbouslog: I also updated the related GDoc iOS - App Store, removing reference to the MARKETING_VERSION in the project.pbxproj file.

PR #424, opened 2022-10-07
Enable dependabot updates

eatyourgreens:
Set up dependabot to run weekly, with a maximum of 10 open PRs at any one time.


mcbouslog (approved):
Thank you for adding!

PR #462, opened 2023-10-09
Adjusted AndroidManifest.xml to 1) Fix android 13 crash and 2) remove…

coryzoon:
Fixes android 13 crash. The MainActivity needed to be set to MAIN instead of LaunchScreen. Removed LaunchScreen (not needed in newer versions and causes issues). Also removed setting for largeHeap. This shouldn't be there unless it's needed as it can cause performance issues.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -11,7 +11,6 @@
 
     <application
       android:name=".MainApplication"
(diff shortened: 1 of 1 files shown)


mcbouslog (approved):
Confirmed app runs locally with Pixel 5, API 33 and Pixel 5, API 31 👍

PR #463, opened 2023-10-09
Update package.json scripts

mcbouslog:
Update package.json scripts to include `react-native`. Current scripts don't run react-native server or react-native builds as expected.

run `npm start` to start react-native local server, `npm android` to build for android, or `npm ios` to build for ios.

- [ ] Are tests passing?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,9 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "start": "npm start",
-    "android": "npm run android",
-    "ios": "npm run ios",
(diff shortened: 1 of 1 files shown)


coryzoon (approved):

coryzoon:
@mcbouslog I've always used `npx react-native start` and when I run `react-native start` I get errors but I've approved the PR because 1) this works for you 2) my issues might be from how I originally installed React Native 3) I have my own aliases that I always use for running these commands, I don't typically use the npm scripts. Just FYI.

PR #465, opened 2023-10-10
Only tag projects with museum mode if user is logged in.

coryzoon:
Disabled museum mode if user is not logged in. This prevents the state where a logged out user clicks on a project and sees 1) no back button 2) generic title 3) different, darker styling. Projects that are marked museum mode will still have these differences if the user is logged in.


mcbouslog (approved):
Noting that staging "mobile drawing project" which was showing the issue of getting stuck in project and unexpected project heading now shows as expected ✅ .

Created a new staging test project with museum mode and used one of my staging users to create Museum role - https://master.pfe-preview.zooniverse.org/lab/1998/collaborators?env=staging and the project does load as expected, in "museum mode," locally where I get stuck in the project and project heading ✅ .

PR #466, opened 2023-10-10
Added property loadMinimal={true} to Swiper fixes the glitch where ea…

coryzoon:
Tutorial slider had a glitch where each time you went to a new slide the positioning would increasingly get misaligned. 

Researching issues in the swiper library I found https://github.com/leecade/react-native-swiper/issues/569#issuecomment-888039704. I applied this property and the issue went away.


mcbouslog (approved):
Confirmed issue on https://www.zooniverse.org/lab/15140/workflows/22326.

Confirmed fix locally with https://master.pfe-preview.zooniverse.org/lab/1974/workflows/3619?env=staging.

LGTM! 👍

PR #467, opened 2023-10-17
Release/v2.10.0

coryzoon:
- Updated app version and build numbers.
- Updated .gitignore to ignore files for upload key and release.
- Updated Podfile.lock and package-lock.json to lock in dependencies.
- Updated bundle identifier in Xcode.
- Corrected path for fonts (build was failing).
- Added loadMinimal={true} to main Swiper file to match bug fix to the other Swiper.
- Updated background color for iOS launch screen.
- Fixed sentry issue where it was dispatching a plain function, not an object.
- Fixed sentry issue where it was calling name on a route that was undefined.


mcbouslog (approved):
Changes LGTM, confirmed local builds on iOS (iPhone 14 Pro iOS 17) and Android (Pixel 5 API 33) 👍 🎉 🚀 !!!

PR #470, opened 2023-10-27
Update eslint parser to show errors in code.

coryzoon:
Errors were not showing in vscode. For example you could remove an import and no error would show. The parser in the eslintrc.js file needed to be updated to the correct parser. Now vscode will show errors like undefined variables and missing imports.


mcbouslog (approved):
Nice catch!

PR #478, opened 2023-11-10
Filter out finished projects.

coryzoon:
The goal is to filter out the finished projects so they are not shown.

There were 2 approaches I tried:
1) Call for projects with separate API calls where state=live & state=paused. Pros: by not calling for finished it saved 1 extra api call in total. Cons: By separating the calls, the order of the projects was different than before.
2) Call for all projects and filter out finished. Pros: order stays the same (except for not showing finished projects). Cons: 1 extra api call.

The API calls are called in parallel so there was no performance gain with method 1 so I oped for method 2 to keep the order of projects the same as before.


mcbouslog (approved):
Makes sense, code looks good. App feels fast to me, performance concerns can be revisited separately (if ever a priority concern). 👍

Issue #357, opened 2020-12-13
Remove "WebView" from the Mobile App Entirely


chelseatroy:
**What are we doing?**

We are removing all of the `WebView` functionality from the app.

**Why are we doing it?**

- Webview, imported directly from react-native as we currently use in the app, is deprecated in lieu of the react native community version. As such, the deprecated one has some issues that won't be fixed. 

**Additional Context**

We initially used Webviews to display projects that were only designed for the browser, back when the app was new and we didn't have many "mobile-friendly" workflows available. That is no longer the case, and we no longer display browser-based projects in the mobile app (removed in [this commit](https://github.com/zooniverse/mobile/commit/6f2159145e711fdaf2f899a54949f78e78a01ec8)). So all views that were designed to show those can be removed. We should also make sure we're not using WebViews anywhere else in the app.

lcjohnso:
Noting this relevant thread in #458 

https://github.com/zooniverse/mobile/pull/458#issuecomment-1747271029
> > @coryzoon FYI re: [0b0e2d7](https://github.com/zooniverse/mobile/commit/0b0e2d75656cfa4ab1f675aad5fa0734e0ced732)
> > Totally OK to bump WebView library, but see #357 -- there shouldn't be any need to keep WebView anymore, as we don't support browser-based classification from within the app.
> 
> @lcjohnso It's currently still in the app being used to "Provide Beta Feedback For This Project". Currently on the production version of the app it is crashing if the user clicks this button. This library update will prevent the crash but I will add that it doesn't actually work properly but at least it won't crash. It attempts to open a link to a google doc and fill out form. It doesn't seem to work so I could possibly just remove the feedback button all together or figure out how to make it work. The google doc link is "https://docs.google.com/forms/d/e/1FAIpQLScq85a1bh9QRgSsKpUOfXB0B-o1y5llJx3AyWhxGyq--LtCrw/viewform?usp=pp_url"
> 
> ![Screenshot 2023-10-04 at 12 33 50 PM](https://user-images.githubusercontent.com/141738821/272648810-3fcf0e7d-76fa-400e-bf77-158bd7535df6.jpeg)


https://github.com/zooniverse/mobile/pull/458#issuecomment-1747332902

> Thanks for the details re: WebView and beta feedback button, that's very helpful. Sorry for sidetracking PR re: this specific topic. Totally OK to move this conversation to #357 or elsewhere if you wish to defer this side conversation.
> 
[...]

PR #484, opened 2023-11-13
Issue 357 remove webview

coryzoon:
Changes based off https://github.com/zooniverse/mobile/issues/357:
The goal of this ticket was to remove all instances of a WebView except one which was used to display the "Field Guide". Any use of WebView should be from the react-native-webview library and not the deprecated React Native WebView.

- Removed WebViewScreen.js component - was used for the "Provide Beta Feedback For This Project" button which is being removed.
- Removed ZooWebView.js component - was used as a fallback when a project had no workflows and no redirect.
- Removed BetaFeedbackView.js component - was decided that this is no longer needed.
- ClassifierContainer.js - removed usage of BetaFeedbackView and any functions used for the component.
- ProjectTitle.js - removed logic where if no workflow and the project had a redirect, it would open the url. Also removed logic where if no workflow, the project would be opened in a WebView.
- PageKeys,js, app.js, RootNavigator.js - removed any references to WebViewScreen.js or ZooWebView.js.


mcbouslog (approved):
Looks good 👍 .

Clicking a link (i.e. within a tutorial step) works as expected (opens my native device's default browser).

PR #502, opened 2024-01-03
Updates swiper index when user swipes or answers question. Prevents t…

coryzoon:
Updates swiper index when user swipes or answers question. Prevents the list from being stuck on index 0 when user leaves and comes back.


mcbouslog (approved):
Nice fix!

PR #507, opened 2024-02-03
Issue 492 ga human readable page names

coryzoon:
- Replaced Firebase/Analytics with Firebase/AnalyticsWithoutAdIdSupport
- Removed FIREBASE_ANALYTICS_COLLECTION_ENABLED false from Info.plist as it was not needed and was preventing screens from being tracked.
- Froze @sentry/react-native to 5.16.0 as the newest version was a breaking change.
- Added screenTracking.js that takes the route name and converts it a to a human friendly name and then sends to GA analytics.

Example of what the new naming convention looks like (MainActivity is from an older version)
Screenshot 2024-02-03 at 6 35 25 AM


src/navigation/screenTracking.js
```diff
@@ -0,0 +1,43 @@
...
+      name = route?.params?.selectedProjectTag
+        ? `Project List - ${titleCase(route.params.selectedProjectTag)}`
+        : 'Project List';
+      break;
+    // As a default convert the route name to title case.
+    default:
+      name = titleCase(route.name);
+  }
+
+  analytics().logScreenView({
```
mcbouslog, line 39: I don't think it's really necessary, but wrapping this (or maybe `gaTrackScreen(newRoute)` in RootNavigator) in a try/catch crossed my mind. For consideration.

mcbouslog (approved):
This will be really interesting to see the results of; nice!

src/navigation/screenTracking.js
coryzoon: Thanks, it might not be needed but if it could fail if the code changes and something other than an object is passed to the function, so I think it's a good idea to add the try/catch so those errors get logged. I've updated the code.

Issue #497, opened 2023-12-13
Android policy compliance improvements: links to account deletion, privacy policy


lcjohnso:
As described in Google's Play Console Policy doc on [User Data](https://support.google.com/googleplay/android-developer/answer/13316080), apps that allow account creation from within the app (we do; internal registration form is provided) are asked to provide a "readily discoverable option to initiate app account deletion from within your app". The current platform-wide mechanism for account deletion is to visit https://panoptes.zooniverse.org/profile and confirm one's intent to delete by entering the current account password (backup option: email contact@zooniverse.org and a team member can delete account on behalf of account holder via admin page).

Google Policy also states that all app should provide "a privacy policy link...within the app itself" (here, the Zooniverse privacy policy: https://www.zooniverse.org/privacy).

I propose the following improvements:
1. Add a link within the app to the Panoptes profile page used for account deletion. Make this link visible to logged-in users, possibly on the Settings page.
2. Add a link to the privacy policy in a highly visible location. Make the link visible to all users, possibly in the navigation menu under the "Connect" options.

I envision the links opening in a default browser (rather than via Webview internal to the app).

@seanmiller26 -- Can you propose placement locations for these two links?

seanmiller26:
@lcjohnso 

Ideally these links shouldn't be too buried, so I have added subsections below 'Connect': 

'About' Privacy link
'Help' Delete my account link

These should be an intuitive place for users to find these specific links.

<img width="375" alt="Screen Shot 2023-12-15 at 4 24 21 PM" src="https://github.com/zooniverse/mobile/assets/89043489/8a022bb2-0e31-4ea4-8174-a4b687802c88">

<img width="375" alt="Screen Shot 2023-12-15 at 4 44 25 PM" src="https://github.com/zooniverse/mobile/assets/89043489/ea33f8ed-9d8a-4d4a-be7a-92ce2a5ba83a">

The additional links (Version, Contact Us, Sign Out/In) are examples of how we can expand these sections in the future.

Preview here (click the side menu) https://www.figma.com/proto/QC931orjhCdGoQogVg7iDZ/Mobile-App?type=design&node-id=2-201&t=uRGS1RPjRZzCbGg0-1&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=2%3A201&mode=design

coryzoon:
@lcjohnso @seanmiller26 A few questions about this ticket:

[...]

PR #508, opened 2024-02-03
Connect zoon icon in nav header to the home screen.

coryzoon:
Per conversation https://github.com/zooniverse/mobile/issues/497#issuecomment-1921789366, the zoon icon in the upper left of the navigation header should be taking the user to the home screen.


mcbouslog (approved):
LGTM!

PR #510, opened 2024-02-08
Issue 393 lose video controls android expand

### LINKED ISSUE #393: bug on Android: Lose video controller options when click 'expand' on video subjects
**What the app is supposed to do:**

When you click on the 'expand' button on a video subject, I would expect that the video controller options (pause, fast forward, etc.) would continue to be available. 

**What the app does instead:**

When you click on the 'expand' button on a video subject, the video controller (pause, fast forward, etc.) goes away/is not available to click on. 

**Exact steps to Reproduce**
1. Go to Marks' https://www.zooniverse.org/lab/15140/collaborators and add yourself as collaborator
2. open that project within 'preview' section of mobile app
3. click 'expand' button on the video subject. 
4. Try to find the video controller. 

**Screenshots**
Screenshot_20220402-103411-582

(Note; ignore the calls&notifications will vibrate note in the screenshot -- that's just something my phone does when I take a screenshot on it - it's not something I see at other times on the screen). 

**(please complete the following information):**
- OS: [e.g. iOS]: Android
- Device: (example:we once had an icon problem that only appeared on the Pixel 2): Moto g power (2021)
- Version [e.g. 2.8.2]:  Android v11

**Additional context**

coryzoon:
Adds react-native-video-controls like (updates package.json, package-lock.json, Podfile.lock)
Updates FullScreenMedia.js to use the react-native-video-controls which adds controls to the react-native-video player.
Adds options to the video player disableFullScreen(the modal already has a button for this), disableBack (doesn't work and they can use the modal close button) and disableVolume (doesn't work and they can control the volume with the phone controls).

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -396,7 +396,7 @@ PODS:
     - React-Core
   - react-native-netinfo (9.5.0):
     - React-Core
-  - react-native-safe-area-context (4.8.2):
+  - react-native-safe-area-context (4.9.0):
     - React-Core
   - react-native-splash-screen (3.3.0):
     - React-Core
@@ -405,7 +405,7 @@ PODS:
     - react-native-video/Video (= 5.2.1)
   - react-native-video/Video (5.2.1):
     - React-Core
-  - react-native-webview (13.6.4):
+  - react-native-webview (13.7.0):
     - RCT-Folly (= 2021.07.22.00)
     - React-Core
   - React-NativeModulesApple (0.72.4):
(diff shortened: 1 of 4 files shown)


mcbouslog (approved):
Changes look good, straightforward package replacement 👍 .

Confirmed locally with emulated Android (Pixel 5 API 33) video works with pause and scrubber.

I'm not sure if it ever worked or an issue with my local Android emulator, but rotating the phone with full screen video did not rotate the video as expected. Not blocking, but maybe something to check/revisit.

I'm 99% certain video subjects have never supported drawing on mobile (and PFE, FEM is the exception where video subjects do support drawing), but noting to double check that @lcjohnso is also under the same impression.

lcjohnso:
> I'm not sure if it ever worked or an issue with my local Android emulator, but rotating the phone with full screen video did not rotate the video as expected. Not blocking, but maybe something to check/revisit.

FYI: I'm OK with no rotation support. But agree that testing that nothing odd / unexpected happens when rotation is enabled would probably be a good idea during pre-release testing.

> I'm 99% certain video subjects have never supported drawing on mobile (and PFE, FEM is the exception where video subjects do support drawing), but noting to double check that @lcjohnso is also under the same impression.

Correct -- we do not support drawing on videos on mobile.

PR #511, opened 2024-02-15
Updated Publications in side drawer to link externally and removed al…

coryzoon:
The goal of this PR is to update "Publications" in the side drawer menu so that it goes to an external link instead of an internal screen. 

Changes include:
Update src/actions/index.js, src/constants/PageKeys.js, and RootNavigator.js to remove any reference to publications.
Remove components Publication.js, PublicationFilter.js, PublicationList.js & constants/publications.js
Update SideDrawerContent.js to point Publications to an external link
Removed any tests or snapshots for the old publications code


mcbouslog (approved):
Love a net -1,233 code change PR! LGTM 🚀

PR #512, opened 2024-02-19
Release/v2.11.0

coryzoon:
Final PR for 2.11.0
Includes version/build bumps plus bug fixes found during release testing.

- Update iOS/Android version/build numbers.
- Fix issue where Karla font had reverted to the old font.
- Updated NotificationLandingPageScreen.js to not wait for projects to go through extractSwipedEnabledProjects() (added time for screen to load becuase it checked workflows which takes time to grab and was not needed for this screen).
- Added scroll view to side drawer menu to support smaller devices now that more content had been added to it.
- Adjusted logo width in side drawer to accomodate smaller devices.
- Removed flex:1 from Settings.js as it was causing issues with scrolling and was not needed.


mcbouslog (approved):
Changes look good, confirmed app works locally on iOS (recently checked Android with different PR). 🎉

PR #514, opened 2024-02-20
Version bumps

coryzoon:
Updated library versions to the versions that dependabot recommends. Tested the mobile app locally to ensure no breaking changes were introduced.


mcbouslog (approved):
iOS and Android looking good locally.

I'm getting some errors opening the Field Guide on Android, but I don't think related to these updates. If you have an actual Android device please check opening a Field Guide and expanding a Field Guide item.

coryzoon:
> iOS and Android looking good locally.
> 
> I'm getting some errors opening the Field Guide on Android, but I don't think related to these updates. If you have an actual Android device please check opening a Field Guide and expanding a Field Guide item.

Thanks Mark, after doing a little digging I was able to reproduce the issue you reported on both Android and iOS. It looks like it's related to the "Mobile Testing - Staging" project. Specifically the issue is line 21 in FieldGuideItemDetail.js where it expects props.icons to be an object but in some cases where the guide does not have icons it would be undefined which causes the error. I'll submit a separate PR for the fix.

PR #568, opened 2024-04-26
Issue 456 auto play multi image

coryzoon:
Description of changes in order of the files listed below:

- Added react-native-modal library for modal.
- Pulls background from projects in order to display in classifier header.
- Update the full screen media component to match designs.
- Updated all classifier screens and any components that are part of the screens to match updated designs.
- Remove different styling when in museum mode (should always be the same regardless of mode). 
- Updated animations to useNativeDrive: false, to remove warnings.
- Added an auto play functionality for multi image classifications. Allows auto play, press and hold to pause, tap to pause, and updated pagination dots.
- Created new components for buttons to support new designs.
- Updated ClassificationPanel.js with the new tabs to match designs.
- Updated Field Guide and Help modals to match new designs.
- Updated styling for Tutorials to match new designs.
- Fixed bug in SizedMarkdown.js where it would sometimes attempt to replace text when there is no text (very rare but worth fixing).
- Updated SafeAreaContainer.js which is a container for the entire app that provides a safe area view with a background color. To match the designs I need to apply logic that would apply different backgrounds depending on if the screen was a classifier screen or not. Preview projects would still have the red background.
- Added ClassifierHeader component for the new headers that would display the blurred project background image. Updated RootNavigator.js to not show the default header so that the custom header could be shown instead.


src/components/classifier/FieldGuide.js
```diff
@@ -12,11 +13,11 @@ import {
 import EStyleSheet from 'react-native-extended-stylesheet'
 import FieldGuideItemDetail from './FieldGuideItemDetail'
 import FieldGuideItemRow from './FieldGuideItemRow'
-import Icon from 'react-native-vector-icons/FontAwesome'
+import Icon from 'react-native-vector-icons/MaterialCommunityIcons'
 import PropTypes from 'prop-types';
-import {addIndex, isEmpty, map} from 'ramda'
+import {isEmpty} from 'ramda'
```
mcbouslog, line 18: Not sure if `ramda` use can be replaced with `lodash` (added in PR 572)

src/components/classifier/SwipeSingleImage.js
```diff
@@ -0,0 +1,46 @@
...
+    elevation: 5,
+  },
+  expandButton: {
+    padding: 16,
+  },
+});
+
+SwipeSingleImage.propTypes = {
+  uri: PropTypes.string,
+  onExpandButtonPressed: PropTypes.bool,
```
mcbouslog, line 43:
```suggestion
  onExpandButtonPressed: PropTypes.func,
```

src/components/classifier/SwipeableSubject.js
```diff
@@ -1,182 +1,189 @@
...
-    Dimensions,
-    Image,
-    ScrollView,
-    TouchableWithoutFeedback,
-    View,
-    Platform
-} from 'react-native';
+import { View, Platform, TouchableOpacity, TouchableWithoutFeedback } from 'react-native';
 import PropTypes from 'prop-types';
 import R from 'ramda';
```
mcbouslog, line 4: Can just import `isEmpty`, or refactor with lodash

src/components/classifier/SwipeableSubject.js
```diff
@@ -1,182 +1,189 @@
...
-    imageUris: PropTypes.array,
-    hasMultipleSubjects: PropTypes.bool,
-    onDisplayImageChange: PropTypes.func,
-}
+  imageUris: PropTypes.array,
+  hasMultipleSubjects: PropTypes.bool,
+  onDisplayImageChange: PropTypes.func,
+  swiping: PropTypes.bool,
+  currentCard: PropTypes.bool,
+  onExpandButtonPressed: PropTypes.bool,
```
mcbouslog, line 144:
```suggestion
  onExpandButtonPressed: PropTypes.func,
```

src/components/classifier/SwipeableSubject.js
```diff
@@ -1,182 +1,189 @@
...
+    this.handleDimensionsChange = this.handleDimensionsChange.bind(this);
+  }
 
-    handleDimensionsChange() {
-        if (this.pager) {
-            setTimeout(() => this.pager.scrollTo({y: 0}), 300)
-        }
+  handleDimensionsChange() {
+    if (this.pager) {
+      setTimeout(() => this.pager.scrollTo({ y: 0 }), 300);
```
mcbouslog, line 30: Not sure if needs to be cleared in `componentWillUnmount`?

mcbouslog (approved):
Great job on a lot of work! I've left some comments but I don't think there's anything that's blocking. The newer (functional) components are excellent, the older legacy components could use some clean up, but I think that's out of scope.

src/components/Markings/DrawingClassifier.js
```diff
@@ -196,12 +196,14 @@ class DrawingClassifier extends Component {
                         />
                     }
                 />
-                <TouchableOpacity disabled={DeviceInfo.isTablet()} onPress={() => this.setState({isModalVisible: true})}
-                                  style={styles.subjectDisplayContainer}>
+                <TouchableOpacity
+                    onPress={() => this.setState({ isModalVisible: true })}
```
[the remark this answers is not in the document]
mcbouslog: Disregard the moving state setting out of render comments, I think that's in a lot of the legacy code.

src/components/classifier/FieldGuide.js
mcbouslog: I see now that `ramda` is used in a lot of different places in a handful of different ways and replacing `ramda` with `lodash` or vice versa is not trivial. Disregard related comments, but maybe add to a someday to look into (if there's overlap between the packages, replacing one with the other).

PR #572, opened 2024-04-29
Add feedback

coryzoon:
This code adds the feedback feature that is currently on the web app, to the mobile app.

A demo of this feature can be found at https://share.vidyard.com/watch/byW7RXopHoafGYKkj8cxhQ? (please note that some styling had been updated since the creation of the demo video but the functionality remained the same):

- Added lodash library.
- Created FeedbackModal.js for the feedback UI.
- Added feedback modal to the QuestionClassifier.js classifier. If workflow has feedback enabled it will show the modal after the user classifies.
- Added feedback.js which holds the logic for retrieving the feedback verbiage. This is mostly copied from web and then altered for mobile.
- Modified how workflows are retrieved since it used to only retrieve workflows with a single task but feedback can have multiple tasks.


seanmiller26:
I'd like to address a small styling question here. Is it possible to limit the width of the modal on wider devices ie. tablets? The modals look great on mobile and even portrait orientation tablets, but when turned to landscape the width seems egregious. 

IMG_0041

src/utils/workflow-utils.js
```diff
@@ -45,9 +45,10 @@ const isValidQuestionWorkflow = (workflow) => {
 
   const questionNotTooLong = firstTask.question.length < 200
   const notTooManyShortcuts = shortcut ? shortcut.answers.length <= 2 : true
-  const doesNotUseFeedback = firstTask.feedback ? !firstTask.feedback.enabled : true;
+  const usesFeedback = firstTask.feedback ? firstTask.feedback.enabled : false;
 
-  if (hasSingleTask && questionNotTooLong && notTooManyShortcuts && doesNotUseFeedback) {
+  // Feedback will have multiple tasks so you chec that the workflow has a single task OR is feedback.
+  if ((hasSingleTask || usesFeedback)&& questionNotTooLong && notTooManyShortcuts) {
```
lcjohnso, line 51: I think the `doesNotUseFeedback`/`usesFeedback` check can be eliminated from `isValidQuestionWorkflow()` because a question task workflow can either have feedback or not and be valid.

The current comment statement is not correct: feedback does not necessitate multiple tasks. Therefore, requiring the `(hasSingleTask || usesFeedback)` logic statement is not necessary.

coryzoon:
> I think pixels would probably be best? The most common portrait iPad width is 1024px. So some number slightly above this seems like a good place to start.

If we limit the size to slightly above the current width then the modal would take up the entire width. I was thinking something like a fixed number i.e. 500px or a percentage like '50%'. I could also apply logic where if the device is a tablet make the modal 50% and if it's a phone keep it how it currently is which is full width minus 16px horizontal margin.

seanmiller26:
> If we limit the size to slightly above the current width then the modal would take up the entire width.

I was thinking we set a max width. So if it ever got to say 1024px, it would not get any wider than that. The reasoning is I like the look of it at full width on a portrait tablet (with margins), but don't like it as much on a landscape tablet.

coryzoon:
> > If we limit the size to slightly above the current width then the modal would take up the entire width.
> 
> I was thinking we set a max width. So if it ever got to say 1024px, it would not get any wider than that. The reasoning is I like the look of it at full width on a portrait tablet (with margins), but don't like it as much on a landscape tablet.

Thanks @seanmiller26 I'll make that change. I'm attaching screenshots of the before and after.
Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-05-15 at 08 52 41
Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-05-15 at 08 52 16

src/utils/workflow-utils.js
lcjohnso: @coryzoon Pinging you here as the PR is getting ready to be closed out.

mcbouslog:
Did a quick scan, apologies if I'm missing something, but is the feedback information saved to the classification metadata?

In the front-end-monorepo there's [this line](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/ClassificationStore.js#L122) that pulls from the [FeedbackStore](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/FeedbackStore.js#L10).

cc: @coryzoon , @lcjohnso

lcjohnso:
Good catch, @mcbouslog. Writing out the feedback results to the feedback key in classification metadata is a high priority, as it's used by Caesar and for other uses. @coryzoon: adding this part of the feedback feature should be included in this release.

The feedback key-value pair written into the classification metadata by PFE looks like this (here, for Classification 568552978 from Planet Hunters at the Adler - Project 22096):
```
'feedback': {'T0': [{
   'id': '12345',
   'answer': '1',
   'success': False,
   'strategy': 'singleAnswerQuestion',
   'failureEnabled': True,
   'failureMessage': "Not quite! This lightcurve shows that the star is  increasing and decreasting in brightness every couple of days in a 'wave-like' pattern. However, there are no dips in the brightness that resemble what we would expect for a transiting planet, as transit dips are generally short (on the order of a couple of hours to a day) and look more like a 'U' shape.",
   'successEnabled': True,
   'successMessage': "Correct! Even though this star shows a 'wave-like' pattern, there are no distinct dips that indicate that there is a transiting planet around this star. "
}]}
``` 
Here is an explanation of those contents:
```
'feedback': {<Workflow Task Key>:
    [ <array of feedback rules (though typically only one is used)>
        { <feedback rule key-value pairs> }
    ]
}
```

mcbouslog (commented):
This is looking good, great to see feedback in the app!

It doesn't serve a functional purpose, but I see that feedback.js are largely helpers from https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/store/feedback/helpers. For maintenance purposes, would it be possible to mirror the file structure, with a code comment including link to the original code? 

@lcjohnso - it's worth considering (not for this release) publishing FEM/lib-classifier to npmjs.org like FEM/lib-react-components so the feedback code can be imported directly. Something to add to the someday pile of Zooniverse efforts.

mcbouslog (approved):
I'll approve this as I've reviewed and looks good, left comments, but nothing blocking. @lcjohnso - can you review classification metadata feedback changes once updated? I'll be out of office this week, returning July 1st.

coryzoon:
> I'll approve this as I've reviewed and looks good, left comments, but nothing blocking. @lcjohnso - can you review classification metadata feedback changes once updated? I'll be out of office this week, returning July 1st.

Most recent commit has this update. Please let me know if anything else needs updated. I tested locally and I'm attaching a screenshot of the classification metadata to show the commit fixes the missing metadata. CC @lcjohnso 
Screenshot 2024-06-24 at 9 23 43 AM

lcjohnso:
Hi @coryzoon -- The feedback metadata is missing the `success` key, which should be a boolean True/False value depending on whether the user-submitted response matched the feedback answer or not (i.e., did the app display a success message or a failure message). This is a required key that should be included.

coryzoon:
> Hi @coryzoon -- The feedback metadata is missing the `success` key, which should be a boolean True/False value depending on whether the user-submitted response matched the feedback answer or not (i.e., did the app display a success message or a failure message). This is a required key that should be included.

Thanks @lcjohnso, that has been updated.

PR #609, opened 2024-07-19
Clear timeout in modal component. Remove shouldAnimateOverlay.

coryzoon:
This bug was a combination of two separate issues and would've only occurred in museum mode. 

1) There was some code from before the reskinning that was no longer needed and could've been removed. It was comparing the swiper index to the subject list to display the subject. At some point the swiper index got higher than the highest value in the subject list and caused the crash.

2) In the feedback modal is a timeout that dismisses the modal after 10 seconds in museum mode. It was not being cleared when it was finished and would increase the swipe index, causing the swipe index to be higher than the length of the subject list and therefore attempting to call the id on a subject list object that didn't exist.

The fix was to 1) remove the code not needed and 2) clear to the timeout during cleanup (standard practice for timeouts).

Update: Also updated version/build numbers since this will get released as a hotfix.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -79,8 +79,8 @@ android {
         applicationId "com.zooniversemobile"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 86
-        versionName "2.12.0"
+        versionCode 87
+        versionName "2.12.1"
     }
     signingConfigs {
         debug {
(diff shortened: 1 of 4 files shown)


mcbouslog (approved):
Code changes look good. 🚀

mcbouslog:
I've also confirmed locally that Planet Hunters at the Adler works as expected logged in as the museum mode user, with an iPad 10th generation, iOS 17.5. I received feedback after clicking Yes/No or swiping left/right, that eventually dismissed if unclicked or went away if clicked Ok or close button.

PR #593, opened 2024-06-11
Release/v2.12.0

coryzoon:
- Bump build/version numbers.
- Add sort=id to workflows API call to fix bug with new workflows not showing up.
- Bump versions of libraries in package.json per dependabot alert suggestions.

PR #618, opened 2024-08-11
useCallback on function, clearTimeout, updated propTypes, add arrow f…

coryzoon:
Addressed feedback from the following PRs (please let me know if you think I missed anything!):

- https://github.com/zooniverse/mobile/pull/568
- https://github.com/zooniverse/mobile/pull/572
- https://github.com/zooniverse/mobile/pull/593

Individual notes for each PR:

https://github.com/zooniverse/mobile/pull/568:
- Not causing an issue, but typically see image imports as static at the top of a file, not sure if this could be.
> I think the way I have it is pretty standard for RN. I couldn't find other examples in RN of it being imported at the top but if you know of one let me know and I can make the update.

- If it'd take more than a minute don't bother, but if the mapStateToProps could be refactored/replaced with useSelector I think it would be preferable/consistent with newer react-redux.
> It would take more than a minute just to test it. I would actually advocate for spending the time to convert it to a functional component and completely remove the need for the mapStateToProps and make the component easier to maintain.

- Unrelated to these changes, but we've added an eventListener on line 68 and I don't think it's removed, for consideration remove the eventListener in componentWillUnmount.
> They deprecated removeEventListener for this version of RN. We would need to update the component to a functional component and then you can unsubscribe in the useEffect.

- onExpandButtonPressed: PropTypes.func,
> Good catch, updated.

- Can just import isEmpty, or refactor with lodash
> Updated 

- Not sure if needs to be cleared in componentWillUnmount?
> Updated 

https://github.com/zooniverse/mobile/pull/572:
- onClose: PropTypes.oneOfType([PropTypes.func, PropTypes.bool]),
> Updated
- useCallback for close()
>  Updated

https://github.com/zooniverse/mobile/pull/593
- Is this also an issue in the MultiAnswerClassifier here ?
> It wasn't causing issues but it should be updated with the arrow function syntax anyway.

### DIFF
--- a/src/components/classifier/FeedbackModal.js
+++ b/src/components/classifier/FeedbackModal.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 import { Modal, View, StyleSheet, TouchableOpacity } from 'react-native';
 
 import Icon from 'react-native-vector-icons/Fontisto';
@@ -15,6 +15,11 @@ const FeedbackModal = ({
 }) => {
   const timeoutRef = useRef(null);
 
+  const close = useCallback(() => {
+    clearTimeout(timeoutRef.current);
+    onClose();
+  }, [onClose]);
+
   useEffect(() => {
     if (inMuseumMode) {
       timeoutRef.current = setTimeout(() => {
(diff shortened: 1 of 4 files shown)


mcbouslog (approved):
Changes LGTM, thank you!

PR #641, opened 2024-10-07
Show video player when tutorial step is mp4.

coryzoon:
- Shows VideoPlayer when tutorial step is an mp4.
- Should match size of tutorial images.
- Should only play when the video tutorial step is active. 
- No functionality with the tutorial images should change.


mcbouslog (approved):
Confirmed locally (Step 2 in https://www.zooniverse.org/lab/1974/tutorial?env=staging). LGTM! 🚀 

Noting these changes are related to the media selected with the "Select Media" button within the tutorial builder. It's out of scope for this PR, but I think if an .mp4 is included in the markdown of the tutorial step builder it will not display.

PR #643, opened 2024-10-25
Added Eras stats to home screen.

coryzoon:
Added API call to eras api.
Removed old circle ribbon.
Added new component for Eras stats.
Updated zooniverse teal color.


mcbouslog (commented):
Code looks good! 👍 

Not sure if a visual loading state of some sort for the new stats request makes sense. I'm not sure how long the request takes for a user with a lot of classification stats. Something for consideration, but not blocking.

I'm having trouble running locally, get the following Render Error:
mobile-render-error

Usually something on my end, going to keep looking into, but posting update.

mcbouslog (approved):
Cleared all the caches, looking good locally! 🎉

PR #644, opened 2024-10-27
Updated dependency versions for dependabot.

coryzoon:
- Updated npm dependency versions to satisfy dependabot.
- Updated  Gemfile.lock to handle the rexml dependabot alert.
- Replaced redux-devtools-extension with @redux-devootls/extension as the original was deprecated and causing issues with upgrading other libraries.
- Updated app.js to work with @redux-devtools/extension and use thunk instead of thunkMiddleware per the upgrade docs.


mcbouslog (approved):
Confirmed app runs as expected (logged in, logged out, navigated, classified) 👍

coryzoon:
Closing, was handled in https://github.com/zooniverse/mobile/pull/644.

PR #658, opened 2024-11-19
Release/v2.13.0

coryzoon:
- Bump version/build numbers.
- Update misspelled conatinerSizer to containerSize.
- Fix bug on multi answer classifier causing classifications to not work.
- Add video support to single question classifier.
- Update tutorial to use react-native-video instead of react-native-video-controls as the latter was causing crashes/Exception java.lang.IndexOutOfBoundsException on android.
- Freeze version for react-native-reanimated/react-native-safe-area-context as they were causing crashes. Will create maintenance tickets to address these later.


mcbouslog (approved):
Tested locally and release is looking good! 👍 
Confirmed noted changes work as expected, bugs fixed, and info updated.

PR #686, opened 2025-06-25
Added i18n, translation files, and updated text to be translated thro…

coryzoon:
### PR notes/considerations
- To use translations at a platform and project level, the code uses namespaces from i18n (https://www.i18next.com/principles/namespaces). You can only have 1 default namespace and you cannot change the language on the non-default namespace (i.e. project) so instead it should add a resource bundle for each language and then use getCurrentProjectLanguage() to choose the bundle language.
- If you choose Português (for example) but the project doesn't have Português then it will default to English. It will still convert the platform text to the preferred language so it might be a mix of English and Português. If we want it so that if Português is not available and therefore the entire page should be English then you need to load the platform language into project bundle and update those texts to pull from the project NS instead of the platform NS. Let me know if you have any questions about this.

### PR changes
- Updated dependencies (includes dependabot updates).
- Replace text with translations in the following places: About.js, DrawableSubject.js, DrawingModeButton.js, ToolNameDrawCount.js, NotificationLandingPageScreen.js, ProjectDisciplines.js, Register.js, SideDrawerContent.js, SignIn.js, ButtonsDrawingModal.js, ClassificationPanel.js, FieldGuide.js, FieldGuideBtn.js, FieldGuideItemDetail.js, FieldGuideItemRow.js, NeedHelpButton.js, SwipeTabs.js, TaskHelpModal.js, Tutorial.js, SubjectLoadingIndicator.js, TranslatoinsLoadingIndicator.js, ProjectList.js, ProjectTitle.js, Settings.js, ClassificationHeader.js
- Updated classifiers to get translations, show loading indicator, load translations.
- Translates the project list if the translation is available. Because this could lead to multiple API calls and because the data does not change often, this list is cached to improve performance.
- In some cases, make button widths more dynamic to support various languages.
- Added new project category for projects that are available in the user's preferred language.
- Added a Language Picker in the settings for the user to select their preferred language.
- Added translations to the disciplines in the global.js file.
- Added redux slice for getting/updating the language settings. This uses the react-native-localize 
- Added a LanguageEffect that ensures the i18n language is updated when the language settings are updated.
- Added languages.js which is the official list of supported languages.
- Added a i18n configuration file that does the following:  1) Imports the translations for each file (right now it just imports each language and loads each language. there might be a more programmatic way of doing this that would lessen the number of lines in this file). 2) Initializes i18n. 3) Adds helper functions for seeing if user's preferred language exists in a project, loading the project translations, changing the platform language, and loading the project list translations

### Future improvements
- Translations are used in various ways depending on if the component is a class or functional component. Over time, it would be nice to convert more class components to functional and utilize hooks as opposed to HOC.
- A few translations were missed in the lokalize order so we'll need a second order at some point in time. The translations that were missed were pretty minor and can requested later once it's determined that no other translations are needed.
- Each language in the locales folder includes 3 files. A js file that came from Panoptes-Front-End, a json file from front-end-monorepo and a mobile json file for the mobile only translations provided by lokalize. There are translations that were left in these files that were never used in the mobile app and never removed from the translations file. A future effort would be to clean these files up and possibly use just one translation file for each language. I attempted to do this but I could tell early on this was going to be an extensive process because of the syntax in a js vs json file. If you converted js files to json you have to wrap the keys/values in double quotes. In some cases the values already had double quotes within the strings and so all of those would need to be escaped.


mcbouslog (approved):
Cursory code review looks good! 🚀 

Tested extensively with TestFlight without issue - great work!! 🌐 

Non-blocking comment for consideration.

PR #703, opened 2025-07-17
Fix bug where question would sometimes not be updated with translations.

coryzoon:
Fixes two bugs found while testing:

- In certain situations the question was not updating when switching between english and the preferred language. It was a race condition and by moving the translation into the Question component versus passing it as a prop, the issue was resolved.
- The app would crash when the user would click on Preview because the preview projects did not have the available_languages property or it was not an array. Fixed by checking that available_languages was an array.


mcbouslog (approved):
Changes LGTM 👍 . 
Main changes are in the `Question` component, conditionally defining `questionTranslated`, and the `ProjectList` component.

PR #740, opened 2025-09-23
Field Guide Fix: check that props has icons before using a potentially null key

### LINKED ISSUE #738: App crashes when field guide opened
**What the app is supposed to do:** show the field guide when the button at the bottom of the classify screen is touched.

**What the app does instead:** For IASC Asteroid Search project, the app crashes.

**Exact steps to Reproduce**
1. Open IASC Asteroid Search project 
2. Click on Field Guide
3. App crashes

**Device Info (confirmed / replicated using):**
- OS: iOS 18.6.2
- Device: 13 Pro
- App Version: 2.14.0/24

**Additional context**
First reported via Freshdesk: https://zooniverse.freshdesk.com/a/tickets/24630

My initial guess: lack of icons may be an issue, but that should be allowed behavior. If this is cause, would a solution be to use placeholders?

coryzoon:
Fix for Field Guide in case that icons aren't present (we not added / uploaded by team): check props before trying to use a potentially null key.

Closes #738 

Invision Mock-ups: 

# Review Checklist

- [ ] Are tests passing?

# My goals for this PR:

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -84,8 +84,8 @@ android {
         applicationId "com.zooniversemobile"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 96
-        versionName "2.14.1"
+        versionCode 97
+        versionName "2.14.2"
     }
     signingConfigs {
         debug {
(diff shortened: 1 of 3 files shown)


mcbouslog (approved):
LGTM!

Issue #771, opened 2025-12-12
Drawing Task - Rectangles shrinking on save


shaunanoordin:
## Functionality Bug?

Original report: [Freshdesk ticket 25406](https://zooniverse.freshdesk.com/a/tickets/25406), dated Mon 8 Dec 2025
> It's seamed like there is a techniqual issue on selecting (drow rectangles) the orders.
>
> The issue occurs after I press on "save and close".
> Am using the Zooniverse's app which is been downloaded from the "Play Store", but it depends on the projects which may needs to open a page or window on the "Edge" or not!!.
> 
> <img width="300" alt="Image" src="https://github.com/user-attachments/assets/acd9dfbf-30fb-4649-9251-c386082ecf4b" /> <img width="300" alt="Image" src="https://github.com/user-attachments/assets/a3952523-1dec-41d1-9989-f3007a0217ae" />

According to the volunteer, when they draw a Rectangle on a Subject image and click "save and close", the drawn rectangle changes shape.

- The expected behaviour: Rectangles should retain the dimensions & position when they were drawn by the volunteer.
- Applies to: possibly all projects with Drawing Tasks that use the Rectangle drawing tool (e.g. GWitchHunters) 

### Dev Notes

- App version 2.15.0/102 on Android 14 (Motorola Moto e15)
- Example screenshots demonstrate the issue in GWitchHunters.

### Status

Initial report. Requires investigation & confirmation.

lcjohnso:
I cannot replicate this issue on iPhone (13 Pro, iOS 18.6.2)

PR #772, opened 2025-12-22
Release/v2.15.1

coryzoon:
### Drawing Classifier Fix (https://github.com/zooniverse/mobile/issues/771)

-   Fixes a bug where rectangles appeared in the wrong position after saving, only when logged in
-   Root cause: When logged in, image dimensions arrived late, but the SVG overlay size was never recalculated
-   Fix: Added componentDidUpdate to MarkableImage.js that recalculates overlay dimensions when subjectDimensions changes

  ### Swiper Library Replacement
  - Replaced react-native-deck-swiper with react-native-swipeable-card-stack
  - The old library had compatibility issues after the React Native upgrade
  - New library uses react-native-reanimated for smoother animations
  - Added custom AnimatedCardWrapper and SwipeOverlay components to maintain the same UX (rotation on swipe, Yes/No labels)

  ### Dependency Updates
  | Package                        | From    | To     |
  |--------------------------------|---------|--------|
  | js-yaml                        | 3.14.1  | 3.14.2 |
  | @react-native/metro-config     | 0.74.89 | 0.82.1 |
  | @react-navigation/drawer       | 6.7.2   | 7.6.0  |
  | react-i18next                  | 15.5.3  | 16.2.0 |
  | markdown-it                    | 13.0.2  | 14.1.0 |
  | react-native-vector-icons      | 10.2.0  | 10.3.0 |
  | i18next                        | 24.2.3  | 25.6.0 |
  | react-native-video             | 5.2.2   | 6.17.0 |
  | react-native-safe-area-context | 4.10.1  | 5.6.1  |
  | @tsconfig/react-native         | 3.0.6   | 3.0.7  |
  | jest                           | 29.7.0  | 30.2.0 |


src/components/classifier/SwipeClassifier.js
```diff
@@ -28,10 +30,92 @@ import ClassifierHeader from '../../navigation/ClassifierHeader';
...
+            Extrapolation.CLAMP
+        );
+        return { opacity };
+    });
+
+    return (
+        <>
+            {/* Yes overlay - appears when swiping right */}
+            <Animated.View style={[overlayStyles.labelContainer, yesStyle]} pointerEvents="none">
+                <Text style={overlayStyles.labelText}>{noLabel}</Text>
```
mcbouslog, line 89: Are the labels switched, should this be the `yesLabel`?

mcbouslog (approved):
Code review looks good, though please see in-line comment regarding swipe labels and edit if necessary.

src/components/classifier/SwipeClassifier.js
mcbouslog: Noting the labels are good/expected, ignore previous comment. Testing locally - swipe to the left = No, swipe to the right = Yes , but the variable names or something here is confusing.

PR #838, opened 2026-06-14
 Multi-task workflows, mini-courses, and project loading performance

coryzoon:
## Summary

  Adds three related improvements to the mobile app:

  - **Multi-task workflows**: supports workflows where volunteers answer multiple tasks before submitting one
  classification.
  - **Mini-courses**: adds periodic educational content during classification.
  - **Project loading performance**: reduces startup API requests and defers project loading until a category or
  project is selected.

  Multi-task workflows and mini-courses already exist in the web app, and the Panoptes API has supported them for
  some time. This PR brings mobile closer to feature parity while substantially changing how projects are loaded.

  ## What Changed

  ### Multi-task: Classifier Refactor

  The four legacy classifier screens each handled their own data loading, layout, and submission. They have been
  replaced with a shared `ClassifierScreen` that renders one of four workflow components:

  - `SingleChoice`
  - `MultiSelect`
  - `Drawing`
  - `Swipe`

  Additional changes include:

  - Added shared hooks for subject queues, workflow resources, project translations, and feedback.
  - Added new drawing components compatible with the unified classifier state.
  - Added async classification submission helpers for the new classifier flow.
  - Retained the legacy classifier screens for rollback purposes, although they are no longer reached through
  normal navigation.

  ### Multi-task: Chain Engine

  - Added helpers for identifying multi-task workflows and determining task progression.
  - Extended classifier state with the current task, task history, and annotations for each task.
  - Added Back navigation that restores previous answers.
  - Changing an earlier answer removes stale downstream answers.
  - Submit button labels change between “Next,” “Done,” and “Submit” based on task position.
  - All task annotations are submitted together in one classification.
  - Single-task workflow submission behavior remains unchanged.
  - The subject queue is managed by the shared classifier so task transitions remain on the same subject.
  - Multi-task Swipe workflows no longer show the next subject underneath the current card.

  ### Mini-course

  - Fetches mini-course content from `tutorials?workflow_id=X&kind=mini-course`.
  - Skips mini-courses for guest users.
  - Adds a modal supporting markdown, inline images, video, opt-out, and restart.
  - Stores mini-course translations in a separate i18next namespace.
  - Tracks signed-in users’ classification counts during the session.
  - Supports fixed or repeating display frequencies.
  - Saves opt-out, progress, and completion state in project preferences using the same data structure as the web
  app.
  - Allows users to restart the mini-course from the classifier panel.

  ### Project Loading Performance

  Previously, opening the home screen loaded the complete mobile project catalog along with workflows, images,
  roles, and translations. This produced many API requests, delayed startup on slower connections, and could
  temporarily show incomplete category results.

  The new loading strategy:

  - Makes the home screen usable without first loading the full project catalog.
  - Loads projects only after the user selects a category.
  - Requests compact project cards instead of full project records.
  - Increases project and workflow page sizes from 20 to 100 to reduce pagination.
  - Loads workflows only for projects in the selected category.
  - Loads full project details only after the user selects a project.
  - Batches project translation requests.
  - Skips project-list translation requests when the selected language is English.
  - Caches category and project results for reuse during the session.
  - Loads notification-related projects independently when the notification screen opens.
  - Prevents partial project lists from appearing while requests are still running.

  ## Notes for Review

  - **HAVI test project bypass:** `src/actions/projects.js` contains temporary handling that allows HAVI project
  `32778` and its supported workflows through the mobile project pipeline despite their current API configuration.
  This should be removed when the backend marks the project as mobile-friendly and standard workflow validation
  supports its multi-task chains.
  - **Legacy classifier screens:** These remain intentionally available for rollback. Removal will be handled
  separately.


mcbouslog (approved):
Cursory code review looks good. Great `/projects` performance improvements, I think they'll make a notable difference!

Please let me know if I'm mistaken with the following, but per the constants in src/actions/projects.js we're limiting the initial multi task project to 32778 and the mini course project to 20616 (with further workflow limitations as defined). That sounds good to me, and code looks good, just confirming my understanding.

Great job with the multi-task and mini-course implementation!! Considerations noted in the PR description are excellent and accurate, and related code looks good.

lcjohnso:
Hi @coryzoon -- One minor comment on the mini course via iOS testing (also submitted via TestFlight feedback function): when closing the minicourse modal, I see a quick flash of the next slide as the modal is closing. Not a big deal, but it's an added bonus if you're able to reproduce and remove.

